PR #5, opened 2021-03-13
Mongo CRUD Rest Implementation

raveen-singh:
## Notion ticket link

[Entity Crud Service Using Mongo](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d?p=0da8e80a3e66489e9ede425827fefc4c)

[Entity Crud REST Endpoints](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d?p=9c78ff129f2a443e808aa692635fb26b)

## Implementation description
* This PR implements the Entity CRUD endpoints with the MongoDb service. 
* Currently still a WIP but is nearing completion. If anyone has any ideas on how to make this better please lmk. 

## Steps to test
1.

* 

## Checklist


backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IEntityService, EntityRequestDTO } from "../interfaces/IEntityService";
+var mongoose = require('mongoose');
+
+
+export class EntityService implements IEntityService {
+
+    /* retrieve the Entity with the given id */
+    async getEntity(id: string): Promise<IEntity> {
+        let entity: IEntity | null;
+        entity = await Entity.findById(mongoose.Types.ObjectId(id));
```
sherryhli, line 12: This should work with the string form of the object ID, just `findById(id)` is fine. We shouldn't need the `require("mongoose")` at the top

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        if (!entity) {
+            throw new Error;
```
sherryhli, line 15: Give an error message, `id ${id} not found`. Also rebase on `main` after #6 gets merged and integrate the logger for error handling

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+        }
+        return entity;
+    }
+
+    /* retrieve all Entities (pagination is nice-to-have future feature) */
+    async getEntities(): Promise<IEntity[]> {
+        return await Entity.find();
```
sherryhli, line 22: Could `find()` throw an error? I think a try-catch is needed

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /* create an Entity with the fields given in the DTO, return created Entity */
+    async createEntity(entity: IEntity): Promise<IEntity> {
+        return await Entity.create(entity);
```
sherryhli, line 27: Try-catch needed, validator failures among other things will throw errors

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+    /* update the Entity with the given id with fields in the DTO, return updated Entity */
+    async updateEntity(id: string, entity: IEntity): Promise<IEntity | null> {
+        return await Entity.findByIdAndUpdate({ _id: mongoose.Types.ObjectId(id) }, entity, { new: true })
```
sherryhli, line 31: `findByIdAndUpdate` doesn't run validators by default, add an exta `runValidators` option

```
findByIdAndUpdate({ _id: mongoose.Types.ObjectId(id) }, entity, { new: true, runValidators: true })
```

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /* delete the entity with the given id */
+    async deleteEntity(id: string): Promise<void> {
+        await Entity.findByIdAndDelete(mongoose.Types.ObjectId(id));
```
sherryhli, line 36: Try-catch and error handling please

backend/typescript/tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig.json to read more about this file */
-
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "ES2020",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
-    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    "target": "ES2020", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
sherryhli, line 6: Can we revert all the changes in this file? Looks like it's just spacing changes

docker-compose.yml
```diff
@@ -29,19 +29,19 @@ services:
...
-      dockerfile: Dockerfile
-    ports:
-      - 5000:5000
-    volumes:
-     - ./backend/python:/app
-    depends_on:
-      - db
-    env_file:
-      - ./.env
+  # py-backend:
```
sherryhli, line 32: Uncomment before merging please

backend/typescript/models/entity.pgmodel.ts
```diff
@@ -0,0 +1,19 @@
+import { Column, Model, Table } from "sequelize-typescript";
+
+@Table({ tableName: "entity" })
+export default class Entity extends Model {
```
sherryhli, line 4: I don't think this is being used since the Postgres `EntityService` implementation isn't done yet. Let's move this to a separate branch

sherryhli (changes requested):
Great work! Did a first pass over the model and service layers, will take a look at the controller after these comments are addressed 🙂

backend/typescript/services/implementations/EntityService.ts
sherryhli: Just realized this isn't too clear, you'll need to wait for the logger to handle errors here (in #6), let me know if it'll help if I separate out the logger into its own PR and get that merged first

backend/typescript/models/entity.mgmodel.ts
```diff
@@ -0,0 +1,35 @@
+import mongoose, { Schema, Document, model } from "mongoose";
+
+export interface IEntity extends Document {
```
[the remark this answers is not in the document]
raveen-singh: Renamed to Entity, there is an _id field that is automatically created by Mongo. What would be the benefits of having 2?

backend/typescript/services/interfaces/IEntityService.ts
```diff
@@ -0,0 +1,26 @@
+import { Entity, IEntity } from "../../models/entity.mgmodel";
+
+enum Options {
+    A, B, C, D,
+}
+
+export interface EntityRequestDTO {
```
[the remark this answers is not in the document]
raveen-singh: Do we have a types file?

backend/typescript/services/implementations/EntityService.ts
raveen-singh: entity.find() is a mongo function that returns null/empty array if no records are found. Would a try-catch still be necessary?

backend/typescript/models/entity.mgmodel.ts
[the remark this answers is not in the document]
sherryhli: Just to assert that it's not nullable. I think we should also just use `id` (string representation of object id) instead of `_id` (actual object id)

Without adding `id` to the interface:
without id

With `id`:
with id

backend/typescript/services/interfaces/IEntityService.ts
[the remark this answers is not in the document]
sherryhli: Yep, I think you need to rebase on `main`

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,51 @@
...
+            throw new Error(`Entity id ${id} not found.`);
+        }
+        return entity;
+    }
+
+    async getEntities(): Promise<Entity[]> {
+        return await MEntity.find();
+    }
+
+    async createEntity(entity: Entity): Promise<Entity> {
```
sherryhli, line 22: Sorry, I overlooked this the first time and also forgot to update in Notion but could we follow the pattern in #6 (in `UserService`)? Argument would be `EntityRequestDTO`, return type would be `Promise<EntityDTO>`.  (the 2 types are the same except `EntityRequestDTO` does **not** contain `id`). Similar changes in the update method as well.

I know it's a lot of repetitive boilerplate and it doesn't make too much of a difference right now, but we want to allow users to have the flexibility to choose which fields from the model to include/exclude in the object passed to/from the controller, and having separate DTOs will allow us to do that.

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,51 @@
...
+        try {
+            const newEntity: Entity | null = await MEntity.create(entity);
+            return newEntity;
+        }
+        catch (e) {
+            throw Error;
```
sherryhli, line 28: We should be throwing the caught error `e` (I know the catch block is useless right now, but wait for the logger!)

Same comment for the update and delete methods.

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,51 @@
...
+            return updatedEntity;
+        }
+        catch (e) {
+            throw Error;
+        }
+    }
+
+    async deleteEntity(id: string): Promise<void> {
+        try {
+            await MEntity.findByIdAndDelete(id);
```
sherryhli, line 43: Let's treat a non-existent id as an error case (when `findByIdAndDelete(id)` returns null), throw an error

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,91 @@
+import express, { Router } from 'express';
+import { Entity } from "../models/entity.mgmodel"
```
sherryhli, line 2: This should not be needed once the service layer uses `EntityRequestDTO`. We don't want the controller directly using the model.

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,91 @@
...
+import EntityService from "../services/implementations/EntityService";
+import { EntityRequestDTO } from "../services/interfaces/IEntityService";
+
+export const entityRouter: Router = Router();
+
+/* Create entity Object */
+entityRouter.post("/", async (_req, res) => {
+    const entityObj: EntityRequestDTO = _req.body.entity;
```
sherryhli, line 10: I think it's just `_req.body`

Also you can remove the underscore, it was previously in the test endpoint to denote that the `req` param isn't being used but that's not the case anymore

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,91 @@
...
+    const IentityObj = {} as Entity;
+    const entService = new EntityService();
```
sherryhli, line 12: 1 instance of `EntityService` can be shared by all the endpoints.

Also just curious, would this approach have worked?
https://www.toptal.com/express-js/nodejs-typescript-rest-api-pt-1

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,91 @@
...
+    IentityObj.stringField = entityObj.stringField;
+    IentityObj.intField = entityObj.intField;
+    IentityObj.enumField = entityObj.enumField;
+    IentityObj.stringArrayField = entityObj.stringArrayField;
+    IentityObj.boolField = entityObj.boolField;
+
+    try {
+        await entService.createEntity(IentityObj);
+        res.send('Success')
```
sherryhli, line 21:
```ts
const entity = await entService.createEntity({
  stringField: req.body.stringField,
  intField: req.body.intField,
  ...,
});
res.status(200).json(entity);
```
I prefer sending the newly created object back (it contains an auto-generated `id` field that the client does not know about), and setting the status code explicitly

Similar changes for the PUT endpoint.

sherryhli (changes requested):
Almost there! A few more changes please

backend/typescript/services/implementations/EntityService.ts
sherryhli: Yeah I think this might be ok for an empty query, but generally `find()` can throw (error arg in the callback) and I'd rather just be safe
https://mongoosejs.com/docs/api.html#model_Model.find

chenrichard10 (approved):

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,51 @@
+import { errorMonitor } from "node:events";
```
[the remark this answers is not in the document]
raveen-singh: I think they must be... they were a few others like this.

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,94 @@
+import MgEntity, { Entity } from "../../models/entity.mgmodel";
+import { IEntityService, EntityRequestDTO, EntityResponseDTO } from "../interfaces/IEntityService";
+
+
+export class EntityService implements IEntityService {
+
+    async getEntity(id: string): Promise<EntityResponseDTO> {
+        let entity: Entity | null;
+        entity = await MgEntity.findById(id);
```
sherryhli, line 9: Wasn't there a try-catch around this before? I think it's still necessary because `findById` can throw if you pass in a malformed id

backend/typescript/services/implementations/EntityService.ts
```diff
@@ -0,0 +1,94 @@
...
+            boolField: newEntity.boolField,
+
+        };
+    }
+    async updateEntity(id: string, entity: EntityRequestDTO): Promise<EntityResponseDTO | null> {
+        let updatedEntity: Entity | null;
+        try {
+            updatedEntity = await MgEntity.findByIdAndUpdate({ _id: id }, entity, { new: true, runValidators: true });
+            if (updatedEntity == null) {
+                throw new Error(`Cannot update entity`);
```
sherryhli, line 69: We can use a more specific error message, this case happens if `id` is not found

backend/typescript/services/interfaces/IEntityService.ts
```diff
@@ -0,0 +1,58 @@
...
+   * @returns the updated Entity
+   */
+    updateEntity(id: string, entity: EntityRequestDTO): Promise<EntityResponseDTO | null>;
+
+    /**
+   * delete the entity with the given id
+   * @param id entity id
+   * @returns nothing
+   */
+    deleteEntity(id: string): void;
```
sherryhli, line 57:
```suggestion
    deleteEntity(id: string): Promise<void>;
```

backend/typescript/rest/entityRoutes.ts
sherryhli: This import is not needed anymore

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,77 @@
...
+entityRouter.post("/", async (req, res) => {
+    try {
+        const newEntity = await entService.createEntity({
+            stringField: req.body.stringField,
+            intField: req.body.intField,
+            enumField: req.body.enumField,
+            stringArrayField: req.body.stringArrayField,
+            boolField: req.body.boolField,
+        });
+        res.status(200).json(newEntity);
```
sherryhli, line 19: Nitpick: status code 201 for created

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    catch (e) {
+        res.status(500).send(e.message);
+    }
+});
+
+/* Update entity object by id */
+entityRouter.put("/:id", async (req, res) => {
+    const id: string = req.params.id;
+    const entityObj: EntityRequestDTO = req.body.entity;
```
sherryhli, line 54: No need to nest data under `entity` as discussed, just do `req.body`

Nitpick: could just follow same pattern as in the POST endpoint, no need to alias this as `entityObj`

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+});
+
+/* Delete entity object by id */
+entityRouter.delete("/:id", async (req, res) => {
+    const id: string = req.params.id;
+
+    try {
+        entService.deleteEntity(id);
+        res.send('Success');
```
sherryhli, line 70:
```suggestion
        res.status(204).send();
```

backend/typescript/rest/entityRoutes.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    catch (e) {
+        res.status(500).send(e.message);
+    }
+});
+
+module.exports = entityRouter
```
sherryhli, line 77: ES6 style export should work, `export default entityRouter;`

backend/typescript/server.ts
```diff
@@ -1,19 +1,24 @@
 import cors from "cors";
 import express from "express";
-
 import { mongo, sequelize } from "./models/index";
-// import PgPerson from "./models/person.pgmodel";
-// import MgPerson from "./models/person.mgmodel";
+import { addOptions } from "sequelize-typescript";
```
sherryhli, line 4: Don't think this is needed

backend/typescript/server.ts
```diff
@@ -1,19 +1,24 @@
...
+var bodyParser = require('body-parser');
+
+const entityRouter = require('./rest/entityRoutes');
```
sherryhli, line 7: ES6 style imports please

backend/typescript/server.ts
```diff
@@ -1,19 +1,24 @@
...
 
 const app = express();
 app.use(cors());
 app.use(express.json());
+app.use(bodyParser.urlencoded({ extended: false }))
+
+
+app.get("/", async (_req, res) => {
+  res.send("Hello!")
+});
```
sherryhli, line 18: Can remove

sherryhli (commented):
A couple more things please:
* Rebase on `main` to get the logger and the top-level types file, and then log all errors in `EntityService`
* Move the Postgres model to your Postgres branch
* Add `**/.DS_Store` to `.gitignore`
* Run the linter with the fix option i.e. `yarn fix` in the `backend/typescript` directory

sherryhli (approved):
🚢

PR #6, opened 2021-03-13
Add TypeScript UserService and AuthService implementations using MongoDB

sherryhli:
## Notion ticket link

* Addresses the user service part of [Firebase authentication & user services using MongoDB (Mongoose)](https://www.notion.so/uwblueprintexecs/Firebase-authentication-user-services-using-MongoDB-Mongoose-1618340d0baa4549a8e38cd521c149c0)
* [Configure TS logger for backend](https://www.notion.so/uwblueprintexecs/Configure-TS-logger-for-backend-f4234bb4329143548896825310127acb)

## Implementation description
* Implemented methods defined in `IUserService`
* Defined MongoDB `User` model using Mongoose
* Integrated Firebase Admin SDK for auth related user management logic
* Set up logger using [winston](https://www.npmjs.com/package/winston), which is useful for keeping track of possible inconsistencies across Firebase and MongoDB user data (caused by rollback failure edge cases)

## Steps to test
1. Generate a Firebase service account private key
    * Go to our project in the [Firebase console](https://console.firebase.google.com)
    * "Project settings" > "Service accounts" > "Generate private key"
    * A file will be downloaded, copy it into `/backend/typescript/` and rename it **`firebaseServiceAccount.json`**
 2. Pull from secret manager **OR** add this line to the root `.env` file in the repo
 ```
 GOOGLE_APPLICATION_CREDENTIALS=firebaseServiceAccount.json
 ```
 3. Comment out the `frontend` and `py-backend` services in `docker-compose.yml`
 4. Run the application:
 ```
docker-compose up --build
```
5. Send requests to the test endpoints in `backend/typescript/server.ts` (note they are all GET requests). Suggested test plan:
    * Create some users by sending requests to `/create-user/:firstName/:lastName`.
    * Update some users sending requests to `/update-user/:userId`
    * Retrieve a user by `userId` by sending a request to `/get-user-by-id/:userId`
    * Retrieve a user by `email` by sending a request to `/get-user-by-id/:email`
    * Retrieve all users by sending a request to `/get-users`
    * Delete a user by `userId` by sending a request to `/delete-user-by-id/:userId`
    * Delete a user by `email` by sending a request to `/delete-user-by-email/:email`

Please check the data in MongoDB Atlas and the Firebase console as well.

## What should reviewers focus on?
* The most important file to look at is `backend/typescript/services/implementations/userService.ts`
* Error handling and "rollbacks": does the logic make sense, are there cleaner ways to do this?
[...]

PR #7, opened 2021-03-15
Add TypeScript AuthService implementation

sherryhli:
## Notion ticket link

Addresses the authentication service part of [Firebase authentication & user services using MongoDB (Mongoose)](https://www.notion.so/uwblueprintexecs/Firebase-authentication-user-services-using-MongoDB-Mongoose-1618340d0baa4549a8e38cd521c149c0)

## Implementation description
* Implemented methods defined in `IAuthService`
* Created a Firebase REST API client as some functionality that we require aren't provided by the Admin SDK (token generation and renewal)
* Note 1: `revokeTokens` only revokes **refresh** tokens, any existing valid access tokens will continue to be valid for the rest of their original lifetimes (1 hour max). A possible future improvement is to add a deny-list for access tokens associated with revoked refresh tokens, but this approach isn't commonly followed because access tokens are meant to be short-lived and the short window when they remain valid is tolerable
* Note 2: For the password reset flow, we can only go as far as generating a password reset link at the moment, we need to implement `EmailService` to complete this feature

## Steps to test
1. Follow steps 1-4 from #6  to set up secrets and run the application
2. Pull secrets from Vault again **OR** add this line to the root `.env`
```
# Retrieve the value from the Firebase console
FIREBASE_WEB_API_KEY=*************************
```
3. Send requests to the test endpoints in `backend/typescript/server.ts` (note they are all GET requests). Suggested test plan:
    * Create a test user if you don't have one using `/create-user/:firstName/:lastName`, the default login email is `firstName@lastName.com` and the password is `s3cur3p455w0rd`. The default user role is "User".
    * Generate an access and refresh token pair by sending a request to `/generate-token/:email/:password`, save the response
    * Check that the user is authorized by sending a request to `/is-authorized/:token/:role`, where `token` is the **access** token in the previous response, and role is the role of your user ("User" by default). Change the role to "Admin" or remove some characters from the token and the endpoint should now return false.
    * Renew your token by sending a request to `/renew-token/:refreshToken`, where refresh token is from the response saved earlier. Save this new response.
    * Revoke your refresh tokens by sending a request to `/revoke-tokens/:userId`, get your `userId` from MongoDB Atlas or use one of the `UserService` test endpoints. Check that revocation was successful by trying `/renew-token/:refreshToken` with the now revoked refresh token (you'll get an exception)
    * Generate a password reset link using `/generate-password-reset-link/:email` and use it to set a new password. Check that reset was successful by trying `/generate-token/:email/:password` with your new password

* `backend/typescript/services/implementations/authService.ts`
* `backend/typescript/utilities/firebaseRestClient.ts`
* How do we feel about this hybrid approach of using both the SDK and the REST API? Could look into using REST only.
* Am I using TS features properly?

## Checklist


backend/typescript/package.json
```diff
@@ -18,6 +18,7 @@
     "express": "^4.17.1",
     "firebase-admin": "^9.5.0",
     "mongoose": "^5.11.17",
+    "node-fetch": "^2.6.1",
```
sherryhli, line 21: `note-fetch` or `axios`?

backend/typescript/server.ts
```diff
@@ -5,11 +5,15 @@ import * as firebaseAdmin from "firebase-admin";
 import { mongo, sequelize } from "./models";
 
 /** ***************************************************************************
- * TEMPORARY: UserService test endpoints
+ * TEMPORARY: UserService and AuthService test endpoints
```
sherryhli, line 8: Test endpoints, can ignore in reviews

backend/typescript/services/implementations/authService.ts
```diff
@@ -0,0 +1,81 @@
...
+  userService: IUserService;
+
+  constructor() {
+    this.userService = new UserService();
+  }
+
+  /* eslint-disable class-methods-use-this */
+  async generateToken(email: string, password: string): Promise<Token> {
+    try {
+      return await FirebaseRestClient.signInWithPassword(email, password);
```
carelynntsai, line 20: Do you need an `await` if you're returning?

sherryhli: I think it's needed in a try-catch so that the catch block will actually execute if the promise is rejected

backend/typescript/server.ts
```diff
@@ -61,7 +65,45 @@ app.get("/delete-user-by-email/:email", async (req, res) => {
...
+app.get("/generate-token/:email/:password", async (req, res) => {
+  const token = await authService.generateToken(
+    req.params.email,
+    req.params.password,
+  );
+  res.status(200).json(token);
+});
+
+app.get("/is-authorized/:token/:role", async (req, res) => {
+  if (req.params.role !== "Admin" && req.params.role !== "User") {
```
alexguo8, line 80: Possible improvement is to use enum for these

backend/typescript/package.json
alexguo8: I think both are ok, so we can keep it as is

alexguo8 (approved):
Looks good to me, left a comment but approving to unblock

backend/typescript/server.ts
xinhaoz: I think role is typed here to be either "Admin" or "User" (correct me if I'm wrong) so we should be fine without the enums

carelynntsai (approved):

backend/typescript/server.ts
zm2he: I believe we do have enums implemented for roles in the user model.  It shouldn't be too difficult to introduce those here. Currently I think in our startercode roles are only "Admin" and "User", but project teams could choose to have more roles.

backend/typescript/server.ts
```diff
@@ -61,7 +65,45 @@ app.get("/delete-user-by-email/:email", async (req, res) => {
   await userService.deleteUserByEmail(req.params.email);
   res.status(204).send();
 });
-/** ************************************************************************ */
+
+const authService: IAuthService = new AuthService();
+
+app.get("/generate-token/:email/:password", async (req, res) => {
```
raveen-singh, line 71: Can we put these routes in a different file and use the express router to keep the server file clean?

xinhaoz (approved):

backend/typescript/server.ts
sherryhli: Will do as part of my next task! These are temporary, just to enable testing

PR #15, opened 2021-04-06
Create GraphQL auth middleware and secure resolvers

sherryhli:
## Notion ticket link

[Integrate authentication service into GraphQL resolvers](https://www.notion.so/uwblueprintexecs/Integrate-authentication-service-into-GraphQL-resolvers-5a4c776a30414197b46fc529e80a4626)

## Implementation description
* Created auth middleware, mirrors the REST version (main difference in function signature), applied to GraphQL schema via the [graphql-middleware](https://www.npmjs.com/package/graphql-middleware) package

## Steps to test
1. Run the application and go to the GraphQL playground at http://localhost:5000/graphql
2. Create a user if you don't have one
3. Generate an access token, this checks that unauthenticated users are able to execute this mutation
```graphql
mutation {
  login(email: "username@domain.org", password: "password")
}
```
4. Refresh token. Again, this checks that unauthenticated users are able to execute this mutation
```graphql
mutation {
  refresh
}
```
5. Try an `entity` query or mutation with an `Authorization` request header (bottom left in the GraphQL playground). Altering this token should result in an `AuthenticationError`
image
6. Try a `user` query or mutation. If you are not an "Admin" you should get an `AuthenticationError`
7. Try the `resetPassword` mutation with a different email than your own, you should get an `AuthenticationError`. Switch to the correct email and you should now get a password reset link
8. Try the `logout` mutation with a different userId than your own, you should get an `AuthenticationError`. Switch to the correct userId and the mutation should execute successfully

cc @alexguo8: I remember you mentioning that GraphQL best practices would have the authorization logic in the business logic layer. This will require some refactoring of our service implementations so I'm planning to put a ticket in the backlog to address it later. I don't think it's a pressing issue as we don't currently have the need to control access at the field level. Let me know your thoughts though 🙂 

## Checklist

### DIFF
--- a/backend/typescript/graphql/index.ts
+++ b/backend/typescript/graphql/index.ts
@@ -1,6 +1,12 @@
 import { makeExecutableSchema, gql } from "apollo-server-express";
+import { applyMiddleware } from "graphql-middleware";
 import { merge } from "lodash";
 
+import {
+  isAuthorizedByEmail,
+  isAuthorizedByRole,
+  isAuthorizedByUserId,
+} from "../middlewares/auth.graphql";
 import authResolvers from "./resolvers/authResolvers";
 import authType from "./types/authType";
 import entityResolvers from "./resolvers/entityResolvers";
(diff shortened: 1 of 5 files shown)


backend/typescript/middlewares/auth.graphql.ts
```diff
@@ -0,0 +1,99 @@
+import { AuthenticationError, ExpressContext } from "apollo-server-express";
```
sherryhli, line 1: This file will be combined with backend/typescript/middlewares/auth.ts (REST version) and annotated with scrubber tags

backend/typescript/graphql/index.ts
```diff
@@ -25,4 +31,29 @@ const executableSchema = makeExecutableSchema({
...
+  Mutation: {
+    createEntity: authorizedByAllRoles(),
+    updateEntity: authorizedByAllRoles(),
+    deleteEntity: authorizedByAllRoles(),
+    createUser: authorizedByAdmin(),
+    updateUser: authorizedByAdmin(),
+    deleteUserById: authorizedByAdmin(),
+    deleteUserByEmail: authorizedByAdmin(),
+    logout: isAuthorizedByUserId("userId"),
+    resetPassword: isAuthorizedByEmail("email"),
```
alexguo8, line 55: Yeah doing it this way instead of in the business logic actually seems to make more sense to me for what we have right now

alexguo8 (approved):

PR #17, opened 2021-04-09
Python Postgres CRUD

chenrichard10:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Entity-CRUD-service-using-PostgreSQL-SQLAlchemy-f9848a4b43f540ef90228fbd02a67484, https://www.notion.so/uwblueprintexecs/Entity-CRUD-REST-endpoints-Flask-089178d3b98140c2bab99f5932a48340)

## Implementation description
* Essentially building off of what was previously made in the pfrp starter code

## Steps to test
1. docker-compose up --build the project, comment out ts-backend portion of the docker-compose.yml
2. Make a GET request to localhost:5000/entities, verify that the output is   ```[ ] ```
3. Make a POST request to localhost:5000/entities, with the JSON as 
  ```
{
    "string_field": "Joe",
    "int_field": 21,
    "string_array_field": ["C"],
    "enum_field": "B",
    "bool_field": true
}
```
Verify that the output matches after making another GET request to localhost:5000/entities and contains an id
4. Make a GET request to localhost:5000/entities/id replacing id with the id of the newly created object. Verify that the object that you recently created is the same.
5. Make a PUT request to localhost:5000/entities/id replacing id with the id of the newly created object, with the newly updated fields as 
  ```
{
    "string_field": "Jane",
    "int_field": 42,
    "string_array_field": ["D"],
    "enum_field": "C",
    "bool_field": false
}
```
Verify that the output matches after making another GET request to localhost:5000/entities/id with the id of the updated object

6. Make a DELETE request to localhost:5000/entities/id with the id of the object. Verify that a GET request to localhost:5000  returns []
7. If all steps work, CRUD functionality for python should be good!

*  Endpoints in routes
*  Validation from entity_dto.py in resources
*  entity.py

## Checklist
- [* ] My PR name is descriptive and in imperative tense
- [* ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [* ] I have run the appropriate linter(s)
- [ *] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/python/app/models/entity.py
```diff
@@ -0,0 +1,44 @@
...
+    # define the entities table
+
+    __tablename__ = 'entities'
+
+    id = db.Column(db.Integer, primary_key=True, nullable=False)
+    string_field = db.Column(db.String, nullable=False)
+    int_field = db.Column(db.Integer)
+    enum_field = db.Column(enum)
+    string_array_field = db.Column(db.ARRAY(db.String))
+    bool_field = db.Column(db.Boolean)
```
sherryhli, line 24: For consistency with our TypeScript-Postgres and TypeScript-Mongo models let's make these all non-nullable

backend/python/app/services/entity_service.py
```diff
@@ -0,0 +1,46 @@
...
+    return updated_entity.to_dict()
+
+
+def delete_entity(id):
+    deleted = Entity.query.filter_by(id=id).delete()
+    db.session.commit()
+    
+    # deleted is the number of rows deleted
+    if deleted == 1:
+        return id
```
sherryhli, line 44: The TypeScript delete entity implementations return void at the moment but I think we'll be changing those to return ID as well so this is ok

backend/python/app/services/entity_service.py
```diff
@@ -0,0 +1,46 @@
+from ..models.entity import Entity
+from ..models import db
+
+
```
sherryhli, line 4: General comment on error handling, I think we should throw an exception when the DB result is None, then the controller layer can catch it and send the appropriate HTTP response. We should also add a logger (can be in separate PR) - I can help look into this

backend/python/app/routes/entity_routes.py
```diff
@@ -0,0 +1,75 @@
+from flask import Blueprint, request
+from flask import jsonify
+
+from ..resources.entity_dto import EntityDTO
+from ..services import entity_service
+
+# defines a shared URL prefix for all routes
+blueprint = Blueprint('entity', __name__, url_prefix='/entities')
+
```
sherryhli, line 9: I know you were following the bootcamp example, but for robustness let's handle error checks with try catch instead of null checks

sherryhli (changes requested):
Looks great overall! Just some small changes around error handling needed. I think there also some references to "restaurants" left in the comments so it'd be nice to remove them

Please also run the formatter (`black`), you can run it through the `py-backend` container:

```
docker-compose up
docker exec -it starter-code-v2_py-backend_1 /bin/bash

# in container now
black .
```

backend/python/app/services/entity_service.py
```diff
@@ -0,0 +1,46 @@
...
+    # Entity is a SQLAlchemy model, we can use convenient methods provided
+    # by SQLAlchemy like query.all() to query the data
+    return [result.to_dict() for result in Entity.query.all()]
+
+
+def get_entity(id):
+    # get queries by the primary key, which is id for the Entity table
+    entity = Entity.query.get(id)
+    if entity is None:
+        raise ValueError
```
sherryhli, line 15: Not sure if `ValueError` is the most fitting error type, I think if `entity` is `None`, it would be because the `id` was not found. I'd have a look at other error types, or raise an `Exception` with a custom message

backend/python/app/services/entity_service.py
```diff
@@ -0,0 +1,46 @@
...
+    return entity.to_dict()
+
+
+def create_entity(entity):
+    new_entity = Entity(**entity)
+    db.session.add(new_entity)
```
sherryhli, line 21: Is it possible for this to raise an exception (what happens if validation on a field fails)? If so the error case should be handled

backend/python/app/routes/entity_routes.py
```diff
@@ -0,0 +1,75 @@
...
+# defines GET endpoint for retrieving a single entity based on a provided id
+@blueprint.route("/<int:id>", methods=["GET"], strict_slashes=False)
+def get_entity(id):
+    try:
+        result = entity_service.get_entity(id)
+    except:
+        error = {"error": "entity not found"}
+        # HTTP status code 404 means Not Found
+        return jsonify(error), 404
```
sherryhli, line 25: Going to deviate a bit from the bootcamp, I think it would make more sense to treat all service layer errors as 500 (server) errors and report the error message exactly as given from the service layer

sherryhli (commented):
Just a few more minor things 🙂 

Could you also have a look at these two PRs: #19 and #20? We decided to follow the TypeScript interface-implementation structure for the Python services as well, so it'd be nice to refactor `EntityService` to be in line with that (either here or in a separate PR is fine). Sorry for the last minute change!

backend/python/app/services/entity_service.py
chenrichard10: We already check this in routes, so I believe it is not possible for it to raise an exception.

backend/python/app/services/implementations/entity_service.py
```diff
@@ -0,0 +1,51 @@
...
+    def get_entities(self):
+        # Entity is a SQLAlchemy model, we can use convenient methods provided
+        # by SQLAlchemy like query.all() to query the data
+        return [result.to_dict() for result in Entity.query.all()]
+
+    def get_entity(self, id):
+        # get queries by the primary key, which is id for the Entity table
+        entity = Entity.query.get(id)
+        if entity is None:
+            raise Exception("Invalid id")
```
sherryhli, line 19: This would be a place to use the logger in addition to raising the exception, `self.logger.error("error message here")`

backend/python/app/services/implementations/entity_service.py
```diff
@@ -0,0 +1,51 @@
...
+        return entity.to_dict()
+
+    def create_entity(self, entity):
+        try:
+            new_entity = Entity(**entity)
+        except Exception as error:
+            raise error
```
sherryhli, line 26: Logging

backend/python/app/services/implementations/entity_service.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return new_entity.to_dict()
+
+    def update_entity(self, id, entity):
+        Entity.query.filter_by(id=id).update(entity)
+        updated_entity = Entity.query.get(id)
+        db.session.commit()
+
+        if updated_entity is None:
+            raise Exception("Invalid id")
```
sherryhli, line 40: Logging

backend/python/app/services/implementations/entity_service.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def delete_entity(self, id):
+        deleted = Entity.query.filter_by(id=id).delete()
+        db.session.commit()
+
+        # deleted is the number of rows deleted
+        if deleted == 1:
+            return id
+
+        raise Exception("Invalid id")
```
sherryhli, line 51: Logging

sherryhli (dismissed):
LGTM, some logging still needs to be done in `EntityService`, up to you if you want to address that here or in a separate PR 🙂

sherryhli (approved):
🚢

PR #18, opened 2021-04-22
Add TypeScript email service

alexguo8:
## Notion ticket link

[Design EmailService interface](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d?p=d680308081414fc4a95bf2a01160e112)

## Implementation description
* Add service for sending out generic emails

## Steps to test
1. Navigate to the Google Cloud Platform console and click on the project/create a new project.
2. Navigate to "APIs & Services" on the left pane and click on Credentials.
image

3. Click "Create Credentials" with OAuth Client ID option, and fill in the form (set user type to internal)
4. Navigate back to the Credentials page and click Create Credentials again with OAuth Client ID option. Select "Web Application" and set "Authorised Redirect URIs" to https://developers.google.com/oauthplayground
image
5. Click Create and keep the Client ID and secret given.

6. Navigate to https://developers.google.com/oauthplayground and click settings. Check off "Use your own OAuth credentials" and fill in the Client ID and secret from Step 6.
image

7. In the box labelled "Input your own scopes" type https://mail.google.com/ and click Authorize APIs. 
image
8. Click the account you want to use for the emails and grant email permissions.

9. Now, click “Exchange authorization code for tokens”.
image
10. Keep these refresh and access tokens for the config file.

11. Create a nodemailer.config.ts file that looks like this and fill in the fields from the steps above.

```
const config: Record<string, any> = {
  service: "gmail",
  auth: {
    type: "OAuth2",
    user: "address@email.com",
    clientId: "",
    clientSecret: "",
    refreshToken: "",
    accessToken: ""
  },
};

export default config;
```
12. Create a new EmailService and call sendEmail
image

* Should the interface for sendEmail include a "from" parameter? Or is it reasonable to assume that they will all be sent from a single email
* Should we add some sample email templates?

## Checklist

### DIFF
--- a/backend/typescript/package.json
+++ b/backend/typescript/package.json
@@ -25,6 +25,7 @@
     "lodash": "^4.17.21",
     "mongoose": "^5.11.17",
     "node-fetch": "^2.6.1",
+    "nodemailer": "^6.5.0",
     "pg": "^8.5.1",
     "reflect-metadata": "^0.1.13",
     "sequelize": "^6.5.0",
--- a/backend/typescript/package.json
+++ b/backend/typescript/package.json
@@ -41,6 +41,7 @@
     "@types/mongoose": "^5.10.3",
     "@types/node": "^14.14.31",
     "@types/node-fetch": "^2.5.8",
+    "@types/nodemailer": "^6.4.1",
     "@types/pg": "^7.14.10",
     "@types/validator": "^13.1.3",
     "@typescript-eslint/eslint-plugin": "^4.4.1",
--- a/backend/typescript/services/implementations/emailService.ts
+++ b/backend/typescript/services/implementations/emailService.ts
@@ -1,0 +1,28 @@
+import nodemailer from "nodemailer";
+import NODEMAILER_CONFIG from "./nodemailer.config";
+import IEmailService from "../interfaces/emailService";
+import Logger from "../../utilities/logger";
+
+const EMAIL_FROM = `"Display Name" <address@email.org>`;
+const transporter = nodemailer.createTransport(NODEMAILER_CONFIG);
+
+class EmailService implements IEmailService {
+  /* eslint-disable class-methods-use-this */
+  async sendEmail(to: string, subject: string, html: string): Promise<void> {
+    const mailOptions = {
+      from: EMAIL_FROM,
+      to,
+      subject,
+      html,
+    };
+
+    try {
+      return await transporter.sendMail(mailOptions);
+    } catch (error) {
+      Logger.error(`Failed to send email. Reason = ${error.message}`);
+      return Promise.reject(error);
+    }
+  }
+}
+
+export default EmailService;
--- a/backend/typescript/services/implementations/emailService.ts
+++ b/backend/typescript/services/implementations/emailService.ts
@@ -1,26 +1,33 @@
-import nodemailer from "nodemailer";
-import NODEMAILER_CONFIG from "./nodemailer.config";
+import nodemailer, { Transporter } from "nodemailer";
 import IEmailService from "../interfaces/emailService";
+import { NodemailerConfig } from "../../types";
 import Logger from "../../utilities/logger";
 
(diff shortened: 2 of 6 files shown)


backend/typescript/services/implementations/emailService.ts
```diff
@@ -0,0 +1,28 @@
...
+import NODEMAILER_CONFIG from "./nodemailer.config";
+import IEmailService from "../interfaces/emailService";
+import Logger from "../../utilities/logger";
+
+const EMAIL_FROM = `"Display Name" <address@email.org>`;
+const transporter = nodemailer.createTransport(NODEMAILER_CONFIG);
+
+class EmailService implements IEmailService {
+  /* eslint-disable class-methods-use-this */
+  async sendEmail(to: string, subject: string, html: string): Promise<void> {
```
sherryhli, line 11: > Should the interface for sendEmail include a "from" parameter? Or is it reasonable to assume that they will all be sent from a single email

Yeah I think it would be nice to allow "from" to be configurable, seems like a reasonable scenario that a project would require sending from multiple email accounts. I suggest adding a constructor to instantiate the transporter, this way users can change the account by using a new instance of `EmailService`

```ts
constructor(nodemailerConfig: NodeMailerConfig) {
    this.transporter = nodemailer.createTransport(nodemailerConfigs);
}
```

backend/typescript/services/implementations/emailService.ts
```diff
@@ -0,0 +1,28 @@
...
+    const mailOptions = {
+      from: EMAIL_FROM,
+      to,
+      subject,
+      html,
```
sherryhli, line 16: Let's rename this variable to `htmlBody` for clarity

backend/typescript/services/implementations/emailService.ts
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    try {
+      return await transporter.sendMail(mailOptions);
+    } catch (error) {
+      Logger.error(`Failed to send email. Reason = ${error.message}`);
+      return Promise.reject(error);
```
sherryhli, line 23: Nit: for consistency with the other services just do `throw error`

backend/typescript/package.json
```diff
@@ -25,6 +25,7 @@
     "lodash": "^4.17.21",
     "mongoose": "^5.11.17",
     "node-fetch": "^2.6.1",
+    "nodemailer": "^6.5.0",
```
sherryhli, line 28: Do we need the type declarations in `@types/nodemailer`?

sherryhli (changes requested):
Great work! One suggestion is to use a separate instance of `transporter` for each `EmailService` instance to allow sending from multiple accounts.

For your other question:
> Should we add some sample email templates?

No, don't think it's necessary

backend/typescript/package.json
alexguo8: Yes, for some reason it didn't complain without it but I will add

backend/typescript/services/implementations/emailService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+class EmailService implements IEmailService {
+  transporter: Transporter;
+
+  constructor(nodemailerConfig: NodemailerConfig) {
+    this.transporter = nodemailer.createTransport(nodemailerConfig);
+  }
+
+  async sendEmail(
+    from: string,
```
sherryhli, line 14: Minor thing, but maybe this should be set in the constructor as well?

sherryhli: Another thought, instead of taking the entire email address as a constructor argument, we can just take the display name since the email address is already in the nodemailer config. This way we have complete control over the format of the from address (don't have the rely on users giving us "Display Name <email address>" exactly)

sherryhli (approved):
🚢

PR #33, opened 2021-04-30
Front-end Create and Update

chenrichard10:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/REST-client-for-Entity-endpoints-57d23ba28155464e8d63a19d977f119e)

Tentative, still fixing some styling issues

## Implementation description
*  Update and Create components with functions similar to Auth

## Steps to test
1. Click Create Button and enter your desired fields
2. Click the submit button
3. Make a get request to localhost:5000/entities and verify that entity was created, grab that ID
4.  Go back to main page and click Update Button, enter desired fields
5. Click submit button
6. Make a get request to localhost:5000/entities and verify that entity was updated

* CrudAPIClient
* CreatePage
* UpdatePage
* Create
* Update

## Checklist
- [* ] I have run the appropriate linter(s)
- [* ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/python/app/rest/entity_routes.py
```diff
@@ -16,7 +16,7 @@
 
 # defines GET endpoint for retrieving all entities
 @blueprint.route("/", methods=["GET"], strict_slashes=False)
-@require_authorization_by_role({"User"})
+# @require_authorization_by_role({"User"})
```
sherryhli, line 19: Uncomment before merging please (there's a fix sitting in #32)

frontend/src/APIClients/CrudAPIClient.ts
```diff
@@ -0,0 +1,35 @@
+import baseAPIClient from "./BaseAPIClient";
+import USER_ACCESS_TOKEN_KEY from "../constants/AuthConstants";
+
```
sherryhli, line 3: Let's rename this file to be `EntityAPIClient.ts`

frontend/src/APIClients/CrudAPIClient.ts
```diff
@@ -0,0 +1,35 @@
...
+const create = async ({ formData }: { formData: any }): Promise<any> => {
```
sherryhli, line 4:
```suggestion
const createEntity = async ({ formData }: { formData: EntityRequest }): Promise<EntityResponse> => {
```
Let's define some types and use them in the function signature. The conversion from `formData` to the request type should happen outside of the API client (in the onSubmit of the component)

```ts
type EntityRequest = {
  stringField: string;
  intField: number;
  ...,
}

type EntityResponse = {
  id: string | number;
  stringField: string;
  intField: number;
  ...,
}
```

frontend/src/APIClients/CrudAPIClient.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    const entityData = JSON.parse(JSON.stringify(formData));
+    delete entityData.id;
+    // should account for it it's for python or typescript
+    const { data } = await baseAPIClient.put(
+      `/entities/${formData.id}`,
+      entityData,
+      { headers: { Authorization: bearerToken } },
+    );
+    console.log(data);
```
sherryhli, line 27: Remove all console.logs please

frontend/src/components/crud/Create.tsx
```diff
@@ -0,0 +1,115 @@
...
+    },
+    bool_field: {
+      type: "boolean",
+      title: "Boolean Field",
+      default: true,
+    },
+  },
+};
+
+// const typescriptSchema: JSONSchema7 = {
```
sherryhli, line 50: Using this schema only should be fine (camelCase). The `BaseAPIClient` will automatically do a casing conversion when making the request if using the Python API 🙂

frontend/src/components/crud/Create.tsx
```diff
@@ -0,0 +1,115 @@
...
+const CreateForm = () => {
+  const [response, setResponse] = useState(false);
+  const [data, setData] = useState("Hello");
+
+  if (response) {
+    console.log("YO");
+    console.log("RESPONSE");
+    console.log(data);
+    return <div> {JSON.stringify(data)} </div>;
+  }
```
sherryhli, line 105: This is ok for testing right now but we'll add a separate page to display the created data.

Nit: the bool response state is not needed, you can just do a null check on data

frontend/src/components/crud/Create.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const onSubmit = async ({ formData }: { formData: any }) => {
+    const result = await CrudAPIClient.create({ formData });
```
sherryhli, line 108: Do the formData to object conversion before calling the API client

frontend/src/components/crud/Update.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    bool_field: {
+      type: "boolean",
+      title: "Boolean Field",
+      default: true,
+    },
+  },
+};
+
+// const typescriptSchema: JSONSchema7 = {
```
sherryhli, line 56: Same comment as in create, using the camelCase schema only should be fine

frontend/src/APIClients/CrudAPIClient.ts
```diff
@@ -0,0 +1,35 @@
...
+  } catch (error) {
+    return error;
+  }
+};
+
+const update = async ({ formData }: { formData: any }): Promise<any> => {
+  const bearerToken = `Bearer ${localStorage.getItem(USER_ACCESS_TOKEN_KEY)}`;
+  try {
+    const entityData = JSON.parse(JSON.stringify(formData));
+    delete entityData.id;
```
sherryhli, line 20: Same comment as above about the function siganture, and this formData to object conversion should happen before this function is called. You can make the id a separate argument to this function so you can reuse the `EntityRequest` type

frontend/src/components/pages/CreatePage.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { useHistory } from "react-router-dom";
+import CreateForm from "../crud/Create";
+
+const MainPageButton = () => {
+  const history = useHistory();
+  const navigateTo = () => history.push("../../");
```
sherryhli, line 7: Does it work if you just push the root, `history.push("/")`?

frontend/src/components/crud/Update.tsx
```diff
@@ -0,0 +1,118 @@
...
+//     },
+//   },
+// };
+
+const UpdateForm = () => {
+  const [response, setResponse] = useState(false);
+  const [data, setData] = useState("Hello");
+
+  if (response) {
+    return <div>{data}</div>;
```
sherryhli, line 107: Needs a stringify to display properly

sherryhli (changes requested):
Looking good! I wouldn't worry about styling too much, we can merge after addressing these PR comments

frontend/src/components/crud/Update.tsx
```diff
@@ -0,0 +1,99 @@
...
+  if (data) {
+    return (
+      <div>
+        <JSONPretty id="json-pretty" data={data} />
+      </div>
+    );
+  }
+
+  const onSubmit = async ({ formData }: { formData: any }) => {
+    const entityData: any = JSON.parse(JSON.stringify(formData));
```
sherryhli, line 91: Why is the JSON parse needed here but not in the create component? Not super important but I think a cleaner way is to convert like this instead of stringifying and then parsing

```ts
let object = {};
formData.forEach((value, key) => object[key] = value);
```

Also would prefer not to type `entityData` as `any` unless absolutely necessary, I think it should be
```ts
EntityRequest & { id: string }
```

frontend/src/components/crud/Create.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const uiSchema = {
+  boolField: {
+    "ui:widget": "select",
+  },
+};
+
+const CreateForm = () => {
+  const [data, setData] = useState<EntityResponse | undefined>(undefined);
```
sherryhli, line 80: Not critical, don't need to fix rn but just a note that I think `null` is a more commonly used default value

frontend/src/components/crud/TableWrapper.tsx
```diff
@@ -0,0 +1,146 @@
...
+  };
+};
+
+const GetTable = (props: any) => {
+  const { data } = props;
+  const columns = React.useMemo(
+    () => [
+      {
+        Header: "id",
+
```
sherryhli, line 49: Why is there a newline here lol, did the linter do this?

sherryhli (commented):
No need to address these comments (I'll handle in my next PR) except for the Python entities PUT endpoint issue and removing the create form id field

frontend/src/components/crud/TableWrapper.tsx
chenrichard10: Yeah

sherryhli (approved):

PR #50, opened 2021-06-03
Add Python REST Validators

emilyslouie:
## Notion ticket link

[Python REST API input validation](https://www.notion.so/uwblueprintexecs/Python-REST-API-input-validation-d372faf0eaa0469cade8e6588f0d4299)

## Implementation description
* Created middleware to ensure validation when using DTO endpoints for the Python backend

## Steps to test
1. Send an invalid DTO shape or invalid type
2. Notice that it returns status 400 with a string that explains the error

* The decorator since this is my first time writing a decorator

## Checklist

### DIFF
--- a/backend/python/app/middlewares/user.py
+++ b/backend/python/app/middlewares/user.py
@@ -4,22 +4,23 @@ from functools import wraps
 from ..resources.create_user_dto import CreateUserDTO
 from ..resources.update_user_dto import UpdateUserDTO
 
-def validate_create_user(clazz):
-    """
-    Determine if request is authorized based on access_token validity and role of client
 
-    :param roles: the set of authorized roles to check for
-    :type roles: {str}
+def validate_request_user(dto_class_name):
+    """
+    Determine if request is valid based on the types of the arguments passed in to create or update a user
+
+    :param dto_class_name: the class name to create or update user dto
+    :type dto_class_name: str
     """
 
     def validate_dto(api_func):
         @wraps(api_func)
         def wrapper(*args, **kwargs):
-            user = eval(clazz)(**request.json)
-            isValid = user.validate()
-            if not isValid:
+            dto = eval(dto_class_name)(**request.json)
+            error_message = dto.validate()
+            if error_message:
                 return (
-                    jsonify({"error": "There is a type error in this request."}),
+                    jsonify({"error": error_message}),
                     400,
                 )
             return api_func(*args, **kwargs)
--- a/backend/python/app/middlewares/user.py
+++ b/backend/python/app/middlewares/user.py
@@ -1,8 +1,5 @@
-from flask import current_app, jsonify, request
+from flask import jsonify, request
 from functools import wraps
-
-from ..resources.create_user_dto import CreateUserDTO
-from ..resources.update_user_dto import UpdateUserDTO
 
 
 def validate_request_user(dto_class_name):
--- a/backend/python/app/middlewares/user.py
+++ b/backend/python/app/middlewares/user.py
@@ -1,0 +1,29 @@
+from flask import current_app, jsonify, request
+from functools import wraps
+
+from ..resources.create_user_dto import CreateUserDTO
+from ..resources.update_user_dto import UpdateUserDTO
+
+def validate_create_user(clazz):
+    """
+    Determine if request is authorized based on access_token validity and role of client
+
+    :param roles: the set of authorized roles to check for
+    :type roles: {str}
+    """
+
+    def validate_dto(api_func):
+        @wraps(api_func)
+        def wrapper(*args, **kwargs):
+            user = eval(clazz)(**request.json)
+            isValid = user.validate()
+            if not isValid:
+                return (
+                    jsonify({"error": "There is a type error in this request."}),
+                    400,
+                )
+            return api_func(*args, **kwargs)
+
+        return wrapper
+
+    return validate_dto
--- a/backend/python/app/middlewares/validate.py
+++ b/backend/python/app/middlewares/validate.py
@@ -2,11 +2,11 @@ from flask import jsonify, request
 from functools import wraps
 
 
-def validate_request_user(dto_class_name):
+def validate_request(dto_class_name):
     """
-    Determine if request is valid based on the types of the arguments passed in to create or update a user
+    Determine if request is valid based on the types of the arguments passed in to create or update a dto
 
-    :param dto_class_name: the class name to create or update user dto
+    :param dto_class_name: the class name to create or update dto
     :type dto_class_name: str
     """
 
--- a/backend/python/app/middlewares/validate.py
+++ b/backend/python/app/middlewares/validate.py
@@ -1,6 +1,16 @@
 from flask import jsonify, request
 from functools import wraps
 
+from ..resources.create_user_dto import CreateUserDTO
+from ..resources.entity_dto import EntityDTO
+from ..resources.update_user_dto import UpdateUserDTO
+
+dtos = {
+    "CreateUserDTO": CreateUserDTO,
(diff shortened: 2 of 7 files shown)


backend/python/app/middlewares/user.py
```diff
@@ -0,0 +1,29 @@
...
+    Determine if request is authorized based on access_token validity and role of client
+
+    :param roles: the set of authorized roles to check for
+    :type roles: {str}
+    """
+
+    def validate_dto(api_func):
+        @wraps(api_func)
+        def wrapper(*args, **kwargs):
+            user = eval(clazz)(**request.json)
```
sherryhli, line 18: Would calling the constructor here fail if `request.json` doesn't contain enough parameters (or if it doesn't contain the correct parameters)?

I really like this design though, very nice and reusable, great work! To make this work the DTO class's constructor likely needs to be modified to take `**kwargs` (keyword arguments). Only the constructor needs to be changed, how the constructor is called will not. For example, `CreateUserDTO`:

```python
# current constructor, this will fail if the not-yet-validated json contains an unexpected field or if its missing a field
def __init__(self, first_name, last_name, email, role, password):
    self.first_name = first_name
    ...

# with kwargs
def __init__(self, **kwargs):
    # if first_name is not provided, self.first_name will be assigned None, will check in validate method
    # kwargs is treated as a dict here
    self.first_name = kwargs.get("first_name")
    ...
```

backend/python/app/resources/create_user_dto.py
```diff
@@ -5,3 +5,16 @@ def __init__(self, first_name, last_name, email, role, password):
         self.email = email
         self.role = role
         self.password = password
+
+    def validate():
+        if self.first_name == null || type(self.first_name) != str:
```
sherryhli, line 10:
```suggestion
        if self.first_name == None || type(self.first_name) != str:
```
Python equivalent is `None` :)

backend/python/app/resources/create_user_dto.py
```diff
@@ -5,3 +5,16 @@ def __init__(self, first_name, last_name, email, role, password):
...
+            return False
+        if self.last_name == null || type(self.last_name) != str:
+            return False
+        if self.email == null || type(self.email) != str:
+            return False
+        if self.role == null || type(self.role) != str:
+            return False
+        if self.password == null || type(self.password) != str:
+            return False
```
sherryhli, line 19: A suggestion is to raise an Exception (with a message indicating which field validation failed), this way we can report those details to the user in the API response

emilyslouie: Hm, I think it might be better to just collect the errors in a list and return it to the jsonify that we return anyways since it's an error we can expect to happen rather than an "exceptional" error if that makes sense. I've added the way I think we could do it in my most recent commit.

backend/python/app/middlewares/user.py
emilyslouie: I've changed it to `first_name=None, last_name=None` in the function's parameters to help give argument hints if someone is trying to use the constructor. Let me know if you still prefer I use `**kwrgs` instead

sherryhli: Yep, I think `kwargs` is still needed just because it's possible for the user to supply unsupported fields, e.g. constructing a `CreateUserDTO` from this request will result in an exception since `foo` is not a constructor argument

```json
{
    "foo": "bar"
}
```

backend/python/app/resources/create_user_dto.py
sherryhli: Ok, sounds good 👍

sherryhli (commented):
Looking good, almost there! Can you validate the create/update entity endpoints as well?

backend/python/app/resources/entity_dto.py
```diff
@@ -23,3 +23,22 @@ def __init__(
         self.string_array_field = string_array_field
         self.enum_field = enum_field.upper() if enum_field is not None else enum_field
         self.bool_field = bool_field
+
```
sherryhli, line 26: Could you modify the `EntityDTO` constructor above to use `kwargs` as well, and remove the previous validations that threw Exceptions (I forgot that they were there)

backend/python/app/resources/entity_dto.py
```diff
@@ -23,3 +23,22 @@ def __init__(
...
+    def validate(self):
+        error_list = []
+        if not self.string_field or type(self.string_field) is not str:
+            error_list.append("The string_field supplied is not a string.")
+        if not self.int_field or not type(self.string_field) is not int:
```
sherryhli, line 31: Typo :)
```suggestion
        if not self.int_field or not type(self.int_field) is not int:
```

backend/python/app/resources/update_user_dto.py
```diff
@@ -1,6 +1,18 @@
 class UpdateUserDTO:
-    def __init__(self, first_name, last_name, email, role):
+    def __init__(self, first_name=None, last_name=None, email=None, role=None):
```
sherryhli, line 2: Please use `kwargs` as in `CreateUserDTO`

backend/python/app/resources/update_user_dto.py
```diff
@@ -1,6 +1,18 @@
...
         self.first_name = first_name
         self.last_name = last_name
         self.email = email
         self.role = role
+
+    def validate(self):
+        error_list = []
+        if not self.first_name or type(self.first_name) is not str:
+            error_list.append("There is an error with the first name in this request.")
```
sherryhli, line 11: Same error message format as the other two classes please

backend/python/app/middlewares/validate.py
```diff
@@ -0,0 +1,27 @@
+from flask import jsonify, request
+from functools import wraps
+
+
```
sherryhli, line 4: Sorry this is my bad but the imports are necessary after all

```python
from ..resources.create_user_dto import CreateUserDTO
from ..resources.entity_dto import EntityDTO
from ..resources.update_user_dto import UpdateUserDTO
```

backend/python/app/middlewares/validate.py
```diff
@@ -0,0 +1,27 @@
...
+    Determine if request is valid based on the types of the arguments passed in to create or update a dto
+
+    :param dto_class_name: the class name to create or update dto
+    :type dto_class_name: str
+    """
+
+    def validate_dto(api_func):
+        @wraps(api_func)
+        def wrapper(*args, **kwargs):
+            dto = eval(dto_class_name)(**request.json)
```
sherryhli, line 16: Did some research and it turns out `eval` should generally be avoided due to security reasons, let's change this line to:
```python
dto = dtos[dto_class_name](**request.json)
```

where `dtos` is a dictionary defined in the same file
```python
dtos = {
    "CreateUserDTO": CreateUserDTO,
    "EntityDTO": EntityDTO,
    "UpdateUserDTO": UpdateUserDTO
}
```

backend/python/app/resources/entity_dto.py
```diff
@@ -1,25 +1,30 @@
...
-        self.string_field = string_field
-        self.int_field = int_field
-        self.string_array_field = string_array_field
-        self.enum_field = enum_field.upper() if enum_field is not None else enum_field
-        self.bool_field = bool_field
+    def validate(self):
+        error_list = []
+        if not self.string_field or type(self.string_field) is not str:
+            error_list.append("The string_field supplied is not a string.")
+        if not self.int_field or not type(self.int_field) is not int:
```
sherryhli, line 17:
```suggestion
        if not self.int_field or type(self.int_field) is not int:
```

sherryhli (commented):
1 last minor thing :)

sherryhli (approved):
Nice, 🚢 it!

PR #54, opened 2021-06-06
Swagger For Python

chenrichard10:
## Notion ticket link

[Set up Swagger UI for REST APIs](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d?p=48d82954b93c4fb8a9760a379a0fa45d)

## Implementation description
Basically the exact same as Alex's PR but in Python. The Swagger json file was renamed to snake_case.

## Steps to test
1. Start the server locally with docker-compose up --build (currently it works with mongo + auth, other combos can be tested after scrubbing) * Note that you may need to change entity_routes.py file to use entity_service_mg
2. Navigate to http://localhost:5000/api-docs
3. Test the login endpoint, then add the access_token using the Authorize button in the top right
4. Test out the remaining endpoints, similar to Postman

* documentation_routes.py
* Testing if endpoints work

## Checklist


backend/python/app/templates/oauth2-redirect.html
```diff
@@ -0,0 +1,75 @@
...
+<body>
+<script>
+    'use strict';
+    function run () {
+        var oauth2 = window.opener.swaggerUIRedirectOauth2;
+        var sentState = oauth2.state;
+        var redirectUrl = oauth2.redirectUrl;
+        var isValid, qp, arr;
+
+        if (/code|token|error/.test(window.location.hash)) {
```
DinuWije, line 15: just curious: what does this if statement do (and what's up with the slashes, like /code)?

sherryhli (commented):
A couple of minor things, otherwise looks good!

backend/python/app/templates/oauth2-redirect.html
sherryhli: I think this file was directly copied over from Swagger's repo so we don't need to look too deeply into it.

Looks like `window.location.hash` returns a part of the URL (see https://www.w3schools.com/jsref/prop_loc_hash.asp), the forward slashes are for denoting a regex pattern I believe

sherryhli (approved):
LGTM, 🚢 it!

PR #55, opened 2021-06-08
Add file name to log messages in TypeScript Backend

zm2he:
## Notion ticket link

[Add file name and line number to log messages in TypeScript backend](https://www.notion.so/uwblueprintexecs/Add-file-name-and-line-number-to-log-messages-in-TypeScript-backend-6d6bbc52cfa64cbbb88c42f28cdd0237)

## Implementation description
* Implemented factory-type wrapper around Winston. For any file, the factory is used to initialize a Winston instance, and the file's name `__filename`, which is a global Node.js object, is passed to it.

## Steps to test
1. Generate errors and info logs, and verify that the file name is displayed at the beginning of the message.
2.  In order to test it, I added an endpoint shown below to test it.
```
entityRouter.get("/test", async (req, res) => {
  lg.error("Flurble");
  res.status(200).json("flurble");
});
```

* Is there any concern that the factory would be inefficient?
* I have kept the previous logger in case it is needed, should it be removed?
* I believe that Winston has an enable stacktrace flag that allows stacktraces to be printed for error messages, we can consider enabling that for error logs; this will display the line numbers.

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/typescript/utilities/logger.ts
```diff
@@ -23,4 +23,31 @@ if (process.env.NODE_ENV !== "production") {
...
+        };
+        return newInfo;
+      })(),
+      winston.format.timestamp(),
+      winston.format.json(),
+    ),
+    transports: [
+      new winston.transports.File({ filename: "error.log", level: "error" }),
+      new winston.transports.File({ filename: "combined.log" }),
+    ],
```
DinuWije, line 43: What do the level, format, and transports do?

zm2he: Transports are the files that you log the data to. Level is the the importance level of the log. For instance, we have normal info logs, and more high priority error logs. More information here: https://github.com/winstonjs/winston

This shouldn't be a big issue since the level & filename configurations were in the logger we had previously.

The format specifies the format the log is outputted in. The only new thing I added was a function to insert the filename into the log.

zm2he:
Let me know if its better to put the file name as a separate json field instead of appending it to the start of the error message.

backend/typescript/utilities/logger.ts
DinuWije: cool, I was just curious about how it worked, thanks!

sherryhli:
> Let me know if its better to put the file name as a separate json field instead of appending it to the start of the error message.

I think this is fine, maybe enclose the file name in square brackets or something similar? e.g. `$[{filename}] ${info.message}`

backend/typescript/utilities/logger.ts
```diff
@@ -23,4 +23,31 @@ if (process.env.NODE_ENV !== "production") {
   );
 }
 
+export const logfactory = (filename: string): winston.Logger => {
```
sherryhli, line 26: > Is there any concern that the factory would be inefficient?

Yeah about this, could we try something similar to this comment: https://github.com/winstonjs/winston/issues/197#issuecomment-10438446 instead. I'd rather not create a new logger for every file if possible

Let's rename the old logger in `logger.ts` to `_logger` and then create a new `logger` like this:
```ts
export const logger = (fileName: string) => {
  return {
    error: (message: string) => {
      _logger.error(`[$fileName}] {message}`);
    },
    info: (message: string) => {
      _logger.info(...);
    }
  }
}
```

Sorry for not bringing this up earlier but hopefully should be a quick change.

sherryhli (changes requested):

tamim-khan (commented):
Another option is to use Child Loggers which lets you add arbitrary metadata tags to each logline
https://github.com/winstonjs/winston#creating-child-loggers

backend/typescript/utilities/logger.ts
zm2he: Factory has been replaced by a function which supplies wrappers for all logger levels.

backend/typescript/utilities/logger.ts
```diff
@@ -13,14 +13,32 @@ const logger: winston.Logger = winston.createLogger({
...
 if (process.env.NODE_ENV !== "production") {
-  logger.add(
-    new winston.transports.Console({
-      format: winston.format.combine(
-        winston.format.timestamp(),
-        winston.format.json(),
-      ),
-    }),
+  _logger.add(
+    new winston.transports.Console(),
```
sherryhli, line 17: The removed format options were redundant right? In the future it'd be nice to make a small PR comment about that 👍

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -4,7 +4,9 @@ import {
   EntityRequestDTO,
   EntityResponseDTO,
 } from "../interfaces/IEntityService";
-import Logger from "../../utilities/logger";
+import logger from "../../utilities/logger";
+
+const Logger = logger(__dirname);
```
sherryhli, line 9:
```suggestion
const Logger = logger(__filename);
```

backend/typescript/services/implementations/emailService.ts
```diff
@@ -1,7 +1,9 @@
 import nodemailer, { Transporter } from "nodemailer";
 import IEmailService from "../interfaces/emailService";
 import { NodemailerConfig } from "../../types";
-import Logger from "../../utilities/logger";
+import logger from "../../utilities/logger";
+
+let Logger = logger(__filename);
```
sherryhli, line 6:
```suggestion
const Logger = logger(__filename);
```

sherryhli (changes requested):
A couple of minor things, please also run `yarn fix` and commit the resulting changes. The only unresolved error will be:
```
/app/utilities/logger.ts
  3:7  error  Variable name `_logger` must match one of the following formats: camelCase, PascalCase, UPPER_CASE  @typescript-eslint/naming-convention
  3:7  error  Unexpected dangling '_' in '_logger'
```

You can either add an ESLint ignore comment or just change the casing of `_logger`

backend/typescript/utilities/logger.ts
zm2he: Yes the format option was redundant, comment added to PR description.

sherryhli (approved):
🚢

sherryhli (approved):

PR #59, opened 2021-06-17
Create TypeScript FileStorageService

emilyslouie:
## Notion ticket link

[Create TS FileStorageService](https://www.notion.so/uwblueprintexecs/Create-TS-FileStorageService-dd9b609e2f36403cb2ebab7688c38e35)

## Implementation description
* Created file storage service for files with create, update, get, and delete function
* Create endpoint only creates if the file name does not already exist, otherwise throws an error

## Steps to test
1. Go to the Firebase console for the project and go to `Storage`
2. Add the file you want to test with to `/backend/typescript`
3. Add `import FileStorageService from "./services/implementations/storageService";` and after `firebaseAdmin.initializeApp()...` add `const file: FileStorageService = new FileStorageService();` to `server.ts`
4. Add the following lines after `const file: FileStorageService = new FileStorageService();` and fill it in with the correct names:
```
// Create file
//console.log(file.createFile('test file name', './testfilepath.txt'));

// Get file
//console.log(file.getFile('test file name'));

// Update file
//console.log(file.updateFile('test file name', './testfilepath.txt'));

// Delete file
//console.log(file.deleteFile('test file name'));
```
5. Start the container with `docker-compose up --build`
6. Uncomment create file method and find that the file has been made in the Firebase dashboard. 
7. Uncomment get file method. Notice you get the signed URL back from the get method, and also notice that the create file method will return error saying the file has already been made.
8. Comment out the create file method. Change the string for the file name in the get file method. Notice that you get an error since there is no file.
9. Comment out the get file method. Uncomment the update file method. Verify that the file is updated in the Firebase dashboard. 
10. Comment out the update file method. Uncomment the delete file method. Verify that the file gets deleted in the Firebase dashboard. 

* Are exceptions handled properly?

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/storageService.ts
+++ b/backend/typescript/services/implementations/storageService.ts
@@ -1,8 +1,10 @@
 import { storage } from "firebase-admin";
+
 import IFileStorageService from "../interfaces/storageService";
 import logger from "../../utilities/logger";
 
 const Logger = logger(__filename);
+
 class FileStorageService implements IFileStorageService {
   bucket: any;
 
--- a/backend/typescript/services/implementations/storageService.ts
+++ b/backend/typescript/services/implementations/storageService.ts
@@ -9,7 +9,7 @@ class FileStorageService implements IFileStorageService {
   bucket: any;
 
   constructor(bucketName: string) {
-    this.bucket = storage().bucket(`gs://${bucketName}}`);
+    this.bucket = storage().bucket(bucketName);
   }
 
   async getFile(
--- a/backend/typescript/services/implementations/storageService.ts
+++ b/backend/typescript/services/implementations/storageService.ts
@@ -6,19 +6,24 @@ const Logger = logger(__filename);
 class FileStorageService implements IFileStorageService {
   bucket: any;
 
-  private expirationTimeMinutes = 60;
-
   constructor() {
     this.bucket = storage().bucket(
       "gs://uw-blueprint-starter-code.appspot.com",
     );
   }
 
-  async getFile(fileName: string): Promise<string> {
+  async getFile(
+    fileName: string,
+    expirationTimeMinutes: number = 60,
+  ): Promise<string> {
+    const expirationDate = new Date();
+    expirationDate.setMinutes(
+      expirationDate.getMinutes() + expirationTimeMinutes,
+    );
     try {
       const res = await this.bucket.file(fileName).getSignedUrl({
         action: "read",
-        expires: Date.now() + this.expirationTimeMinutes * 1000,
+        expires: expirationDate,
       });
       return res[0];
     } catch (error) {
--- a/backend/typescript/services/implementations/storageService.ts
+++ b/backend/typescript/services/implementations/storageService.ts
@@ -8,10 +8,8 @@ const Logger = logger(__filename);
 class FileStorageService implements IFileStorageService {
   bucket: any;
 
-  constructor() {
-    this.bucket = storage().bucket(
-      "gs://uw-blueprint-starter-code.appspot.com",
-    );
+  constructor(bucketName: string) {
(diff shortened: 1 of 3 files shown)


sherryhli (commented):
Looks good! Just a couple of small things to address

backend/typescript/services/implementations/storageService.ts
```diff
@@ -0,0 +1,79 @@
...
+        expires: expirationDate,
+      });
+      return res[0];
+    } catch (error) {
+      Logger.error(`Failed to get entity. Reason = ${error.message}`);
+      throw error;
+    }
+  }
+
+  async createFile(fileName: string, filePath: string): Promise<void> {
```
sherryhli, line 35: As mentioned in #62, can we add a `contentType` parameter for callers to specify the file's MIME type?

backend/typescript/services/implementations/storageService.ts
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      const currentBlob = await this.bucket.file(fileName);
+      if ((await currentBlob.exists())[0]) {
+        throw new Error(`File name ${fileName} already exists`);
+      }
+      await this.bucket.upload(filePath, {
+        destination: fileName,
+      });
```
sherryhli, line 43: I think you'll need to indicate content type through the `metadata` field

ref: https://googleapis.dev/nodejs/storage/latest/Bucket.html#upload, https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON

Probably this, check the syntax though
```suggestion
      await this.bucket.upload(filePath, {
        destination: fileName,
        metadata: { contentType }
      });
```

backend/typescript/services/implementations/storageService.ts
```diff
@@ -0,0 +1,79 @@
...
+    } catch (error) {
+      Logger.error(`Failed to upload file. Reason = ${error.message}`);
+      throw error;
+    }
+  }
+
+  async updateFile(fileName: string, filePath: string): Promise<void> {
```
sherryhli, line 50: Same comment about content type

sherryhli (changes requested):

backend/typescript/services/implementations/storageService.ts
```diff
@@ -0,0 +1,89 @@
...
+import logger from "../../utilities/logger";
+
+const Logger = logger(__filename);
+
+class FileStorageService implements IFileStorageService {
+  bucket: any;
+
+  constructor(bucketName: string) {
+    this.bucket = storage().bucket(`gs://${bucketName}}`);
+  }
```
tamim-khan, line 13: Made `bucketName` a parameter to allow using multiple `FileStorageService` instances with different bucket names. Initializing `firebase-admin` with a bucket name in `server.ts` is not required.

backend/typescript/services/implementations/storageService.ts
```diff
@@ -0,0 +1,89 @@
+import { storage } from "firebase-admin";
+
+import IFileStorageService from "../interfaces/storageService";
+import logger from "../../utilities/logger";
+
+const Logger = logger(__filename);
+
+class FileStorageService implements IFileStorageService {
+  bucket: any;
```
sherryhli, line 9:
```suggestion
  bucket: string;
```

sherryhli (approved):
LGTM, 🚢 it!

backend/typescript/services/implementations/storageService.ts
tamim-khan: It's an object, not a string. We need to add a dependency to strongly type it.

sherryhli: Ah gotcha, we can leave it like this for now then

PR #62, opened 2021-06-23
Create Python FileStorageService

alexguo8:
## Notion ticket link

[Create Python FileStorageService](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d?p=92d1b47901da4b128aff4285bd5da57b)

## Implementation description
* Added a file storage service for blobs with create, update, get, and delete functions
* Create endpoint only creates if file name doesn't already exist, if it does we throw an exception
* Update endpoint only updates if file name exists, if not we throw an exception

## Steps to test
1. Visit the Firebase console for this project and navigate to Storage
image
2. Add a file you want to test with to your /app directory

3. Add the DEFAULT_BUCKET env value. You may also want to change `erase_db_and_sync` to be False in models/\_\_init\_\_.py so that the terminal isn't bloated when running

4. Add the following lines of code to the root \_\_init\_\_.py

```
from firebase_admin import credentials
from flask import current_app
from .services.implementations.file_storage_service import FileStorageService
```

After `rest.init_app(app)` add the following lines of code
```
    file_storage_service = FileStorageService(current_app.logger)
    script_dir = os.path.dirname(__file__)
    rel_path = "your_file.ext"
    abs_file_path = os.path.join(script_dir, rel_path)

    # Get File
    # print(file_storage_service.get_file("your_file.ext"))

    # Create File
    # with open(abs_file_path, 'rb') as file:
    #    print(file_storage_service.create_file('your_file.ext', file))

    # Update File
    # with open(abs_file_path, 'rb') as file:
    #    print(file_storage_service.update_file('your_file.ext', file))
    
    # Delete File
    # file_storage_service.delete_file("your_file.ext")
```
5. Start the container with docker-compose up --build
6. Uncomment Create File and verify that your file is created in the Firebase dashboard. Also verify that you get the signed url to the file. Run it again and verify that you get an exception. 
7. Comment Create File and uncomment Get File. Verify that the signed url to the file is printed
8. Comment Get File and uncomment Update File. Verify that your file is updated in the Firebase dashboard. Also verify that you get the signed url to the file. Change the string argument to another file, and verify that you get an exception.
9. Comment Update File and uncomment Delete File. Verify that your file is deleted in the Firebase dashboard. Change the string argument to another file and verify that you get an exception
10. Comment Delete File and uncomment Get File. Verify that None is printed

* Are we handling exceptions properly?
* Does the behaviour of create and update make sense? An alternative is that we could have a single function that always either creates or replaces a file without any checks

## Checklist


backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,80 @@
...
+            reason = getattr(e, "message", None)
+            self.logger.error(
+                "Failed to update file {name}. Reason = {reason}".format(
+                    name=file_name, reason=(reason if reason else str(e))
+                )
+            )
+            raise e
+        expiration = datetime.now() + self.expiration_time
+        url = blob.generate_signed_url(expiration)
+        return url
```
tamim-khan, line 68: Should update return a URL to the file?

alexguo8: I wasn't sure about this, open to other suggestions
Do you think it should return the blob instead? Or just nothing

tamim-khan: I was thinking `create_file` and `update_file` should not return anything

backend/python/app/__init__.py
```diff
@@ -50,7 +51,10 @@ def create_app(config_name):
     app.config["MONGODB_URL"] = os.getenv("MG_DATABASE_URL")
 
     # required for auth
-    firebase_admin.initialize_app()
+    cred = credentials.Certificate("./firebaseServiceAccount.json")
+    firebase_admin.initialize_app(
+        cred, {"storageBucket": os.getenv("DEFAULT_BUCKET")}
```
sherryhli, line 56: Minor thing, but is it possible to initialize this without `cred`? The path to the private key (./firebaseServiceAccount.json) is already set as an env variable so it should automatically load the credentials

backend/python/app/services/interfaces/file_storage_service.py
```diff
@@ -0,0 +1,60 @@
+from abc import ABC, abstractclassmethod, abstractmethod
```
sherryhli, line 1:
```suggestion
from abc import ABC, abstractmethod
```

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, logger):
+        """
+        Create an instance of FileStorageService with default bucket
+
+        :param logger: application's logger instance
+        :type logger: logger
+        """
+        self.logger = logger
+        self.expiration_time = timedelta(minutes=60)
```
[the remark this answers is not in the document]
sherryhli: > Another option could be to have the expiration_time as an optional parameter to get_file with a default value of 60 minutes. I would prefer this approach if update_file does not need to return a signed URL

I like this suggestion, could make expiration time configurable for create and update as well 👍

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,80 @@
...
+            reason = getattr(e, "message", None)
+            self.logger.error(
+                "Failed to create file {name}. Reason = {reason}".format(
+                    name=file_name, reason=(reason if reason else str(e))
+                )
+            )
+            raise e
+        expiration = datetime.now() + self.expiration_time
+        url = blob.generate_signed_url(expiration)
+        return url
```
sherryhli, line 49: I think it's still a good idea to return a link, as all of our POST and PUT endpoints return the entire created/updated object. Thoughts?

sherryhli (commented):
Looks good! I think we need another quick discussion about returning a url for create and update but otherwise good to go

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,73 @@
+import firebase_admin
```
sherryhli, line 1: Unused import

```suggestion
```

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_file(self, file_name, expiration_time=timedelta(minutes=60)):
+        blob = self.bucket.get_blob(file_name)
+        if not blob:
+            return None
+        expiration = datetime.now() + expiration_time
+        url = blob.generate_signed_url(expiration)
+        return url
+
+    def create_file(self, file_name, file):
```
sherryhli, line 31: Can we add a `content_type` parameter for callers to specify MIME type (e.g. `application/pdf`)? Otherwise it will default to `application/octet-stream` in Firebase Storage which is "arbitrary binary data".

cc @emilyslouie @tamim-khan

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,73 @@
...
+        except Exception as e:
+            reason = getattr(e, "message", None)
+            self.logger.error(
+                "Failed to create file {name}. Reason = {reason}".format(
+                    name=file_name, reason=(reason if reason else str(e))
+                )
+            )
+            raise e
+
+    def update_file(self, file_name, file):
```
sherryhli, line 47: Same comment about content type

backend/python/app/services/interfaces/file_storage_service.py
```diff
@@ -0,0 +1,60 @@
+from abc import ABC, abstractmethod
+
+
+class IFileStorageService(ABC):
+    """
+    FileStorageService interface for handling blob storage
+    """
+
+    @abstractmethod
+    def get_file(self, file_name):
```
sherryhli, line 10: Update signature with `expiration_time` please, and document that the expected type is a `timedelta`

backend/python/app/services/interfaces/file_storage_service.py
```diff
@@ -0,0 +1,60 @@
...
+    def create_file(self, file_name, file):
+        """
+        Creates a file in the default bucket with given file name
+
+        :param file_name: name of the file
+        :type file_name: str
+        :param file: file to upload
+        :type file: file
+        :return: signed url of the created file
+        :rtype: str
```
sherryhli, line 31: Please remove

backend/python/app/services/interfaces/file_storage_service.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Replaces the file that has given file name in default bucket
+
+        :param file_name: name of the file
+        :type file_name: str
+        :param file: file to replace with
+        :type file: file
+        :return: signed url of the updated file
+        :rtype: str
+        :raises Exception: file name does not already exist
```
sherryhli, line 47: Please remove

backend/python/app/services/implementations/file_storage_service.py
```diff
@@ -0,0 +1,73 @@
...
+        url = blob.generate_signed_url(expiration)
+        return url
+
+    def create_file(self, file_name, file):
+        current_blob = self.bucket.get_blob(file_name)
+        if current_blob:
+            raise Exception("File name {name} already exists".format(name=file_name))
+        blob = self.bucket.blob(file_name)
+        try:
+            blob.upload_from_file(file)
```
sherryhli, line 37:
```suggestion
            blob.upload_from_file(file, content_type=content_type)
```

ref: https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.upload_from_file

sherryhli (changes requested):
Sorry didn't catch this last time but can we expand `create` and `update` to take a `content_type` parameter?

sherryhli (approved):
LGTM, 🚢 it!

FYI code freeze into `main` is over now so feel free to merge (commits part of the last release have already been merged into a release feature branch)

PR #63, opened 2021-06-28
Fix several minor issues

sherryhli:
## Notion ticket link

N/A - fixing a couple of small issues that came up when testing the release branch
* Resolve `body-parser` deprecation warning message
* Fix casing issue in Python backend's openapi.json, `userId` -> `user_id`
* Allow empty list to be supplied as entity's `string_array_field` and simplify validation checks in Python backend
* Add newline to Logout.tsx (lint failure was causing frontend compile to fail)

## Implementation description
* N/A

## Steps to test
1. af26a93: Run the TypeScript backend, verify that this deprecation warning is not logged
```
body-parser deprecated undefined extended: provide extended option
```
2. 56575c9: Run the Python backend and navigate to `/api-docs`, try a GET user request with a `user_id` query parameter and a DELETE user request with a `user_id` query parameter
3. 286fa7b: Using the Python backend, try POST and PUT entities and users requests with missing or improperly typed fields and verify that 400 status codes are returned
4. 66a2797: Trivial (tested when creating the release already)
5. acf63c7: Renamed TS storageService.ts to fileStorageService.ts for consistency with Python backend

* Review commit by commit, each fixes a different issue

## Checklist


tamim-khan (approved):
LGTM

backend/python/app/static/openapi.json
```diff
@@ -339,7 +339,7 @@
         ],
         "description": "Delete an entity based on its ID",
         "responses": {
-          "204": {
+          "200": {
```
alexguo8, line 342: Maybe we should change delete entity endpoint entity_routes.py to return 204 instead of changing this, as that's what Python user/auth and Typescript user/auth/entity delete endpoints return

sherryhli: Yeah we are inconsistent on whether we return content or not from DELETE endpoints across the different endpoints. It's pretty typical to not return anything in REST APIs so nothing wrong with changing this to be a 204, but I've read that it's best practice to always return something from GraphQL APIs, so for consistency I think we may actually change all REST endpoints to return something (the deleted id) as well.

Let's merge this change so at least the documentation reflects the actual API behaviour right now, but we can have a discussion sometime about standardizing API return types/statuses

alexguo8 (approved):

tamim-khan (approved):

PR #86, opened 2021-07-07
Update EntityService Mongo

### LINKED ISSUE #81: Integrate Python FileStorageService into EntityService

tamim-khan:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Task-Board-db95cd7b93f245f78ee85e3a8a6a316d)

Closes #81

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/python/app/resources/entity_dto.py
+++ b/backend/python/app/resources/entity_dto.py
@@ -8,6 +8,7 @@ class EntityDTO(object):
             if kwargs.get("enum_field") is not None
             else kwargs.get("enum_field")
         )
+        self.file_name = kwargs.get("file_name")
         self.bool_field = kwargs.get("bool_field")
 
     def validate(self):
--- a/backend/python/app/rest/entity_routes.py
+++ b/backend/python/app/rest/entity_routes.py
@@ -8,9 +8,10 @@ from ..middlewares.validate import validate_request
 
 # from ..services.implementations.entity_service import EntityService
 from ..services.implementations.entity_service_mg import EntityService
+from ..services.implementations.file_storage_service import FileStorageService
 
 # define instance of EntityService
-entity_service = EntityService(current_app.logger)
+entity_service = EntityService(current_app.logger, FileStorageService(current_app.logger))
 
 # defines a shared URL prefix for all routes
 blueprint = Blueprint("entity", __name__, url_prefix="/entities")
--- a/backend/python/app/services/implementations/entity_service_mg.py
+++ b/backend/python/app/services/implementations/entity_service_mg.py
@@ -1,10 +1,12 @@
+from ..interfaces.file_storage_service import IFileStorageService
 from ...models.entity_mg import Entity
 from ..interfaces.entity_service import IEntityService
 
 
 class EntityService(IEntityService):
-    def __init__(self, logger):
+    def __init__(self, logger, file_storage_service: IFileStorageService):
         self.logger = logger
+        self.file_storage_service = file_storage_service
 
     def get_entities(self):
         entities = []
@@ -24,8 +26,9 @@ class EntityService(IEntityService):
 
         return entity.to_serializable_dict()
 
-    def create_entity(self, entity):
+    def create_entity(self, entity, file, content_type):
         try:
+            self.file_storage_service.create_file(entity.file_name, file, content_type)
             new_entity = Entity(**entity.__dict__)
             new_entity.save()
         except Exception as error:
@@ -34,7 +37,9 @@ class EntityService(IEntityService):
 
         return new_entity.to_serializable_dict()
 
-    def update_entity(self, id, entity):
+    def update_entity(self, id, entity, file, content_type):
+        if file:
+            self.file_storage_service.update_file(entity.file_name, file, content_type)
         updated_entity = Entity.objects(id=id).modify(new=True, **entity.__dict__)
 
         if updated_entity is None:
@@ -45,7 +50,10 @@ class EntityService(IEntityService):
 
     def delete_entity(self, id):
         try:
+            file_name = Entity.objects.get(id=id).file_name
             deleted = Entity.objects(id=id).modify(remove=True)
+            if deleted:
+                self.file_storage_service.delete_file(file_name)
             return id
         except Exception as error:
             self.logger.error(str(error))


backend/python/app/services/implementations/entity_service_mg.py
```diff
@@ -45,7 +50,10 @@ def update_entity(self, id, entity):
 
     def delete_entity(self, id):
         try:
+            file_name = Entity.objects.get(id=id).file_name
```
zm2he, line 53: Mongo should return the entity when deleted, so you shouldn't need to make an extra query.

backend/python/app/rest/entity_routes.py
```diff
@@ -135,3 +154,16 @@ def delete_entity(id):
...
 
     return jsonify(result), 200
+
+
+# defines GET endpoint for a URL to the entity's file with the provided uuid
+@blueprint.route("/files/<string:id>", methods=["GET"], strict_slashes=False)
+@require_authorization_by_role({"user", "Admin"})
+def get_file(id):
+    try:
+        file_url = FileStorageService.get_file(id)
```
tamim-khan, line 164: @zm2he `get_file` is not a static method

zm2he: Made file storage service an instance variable in routes before passing it to entity service. Will test to see if it works.

sherryhli (changes requested):
Looking good! Main change would be to remove `file_name` as an `EntityDTO` field. Doesn't make a practical difference but it would align with the TS implementation, also personal preference :)

sherryhli:
@alexguo8 can you take a quick pass at this as well to make sure it aligns with the TS implementation?

backend/python/app/rest/entity_routes.py
```diff
@@ -135,3 +157,16 @@ def delete_entity(id):
         return jsonify({"error": (error_message if error_message else str(e))}), 500
 
     return jsonify(result), 200
+
+
+# defines GET endpoint for a URL to the entity's file with the provided uuid
+@blueprint.route("/files/<string:id>", methods=["GET"], strict_slashes=False)
+@require_authorization_by_role({"user", "Admin"})
```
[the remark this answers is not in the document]
zm2he: Fixed.

backend/python/app/resources/entity_dto.py
```diff
@@ -1,3 +1,12 @@
+allowable_content_types = [
```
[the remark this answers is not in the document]
zm2he: Fixed.

alexguo8:
Aside from the DTO changes Sherry mentioned, lgtm
In TypeScript we had to change FileStorageService to have a bucketName field instead of bucket since FileStorageService is created at the top level and we couldn't initialize the bucket before firebase initializeApp() was called
Was this not an issue for Python? Not sure if this needs to be changed to be consistent @sherryhli

sherryhli:
> In TypeScript we had to change FileStorageService to have a bucketName field instead of bucket since FileStorageService is created at the top level and we couldn't initialize the bucket before firebase initializeApp() was called
Was this not an issue for Python? Not sure if this needs to be changed to be consistent @sherryhli

I suspect this isn't an issue in the Python backend because we're synchronously initializing firebase and then `FileStorageService`. I'm okay with letting this differ from the TS backend for now. The better solution might be checking to see if we can modify TS to synchronously initialize as well (will put in a ticket for next term).

alexguo8:
> > In TypeScript we had to change FileStorageService to have a bucketName field instead of bucket since FileStorageService is created at the top level and we couldn't initialize the bucket before firebase initializeApp() was called
> > Was this not an issue for Python? Not sure if this needs to be changed to be consistent @sherryhli
> 
> I suspect this isn't an issue in the Python backend because we're synchronously initializing firebase and then `FileStorageService`. I'm okay with letting this differ from the TS backend for now. The better solution might be checking to see if we can modify TS to synchronously initialize as well (will put in a ticket for next term).

Oh I see, we only import the routes files after we call firebase_admin.initialize_app, ok that makes sense 👍

backend/python/app/services/implementations/entity_service.py
```diff
@@ -22,6 +25,13 @@ def get_entity(self, id):
 
     def create_entity(self, entity):
         try:
+            if entity.file:
+                entity.file_name = str(uuid4())
```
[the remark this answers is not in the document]
zm2he: Fixed.

backend/python/app/services/implementations/entity_service.py
```diff
@@ -34,6 +44,26 @@ def create_entity(self, entity):
...
+            self.logger.error("Invalid id")
+            raise Exception("Invalid id")
+
+        if entity.file:
+            if current_entity.file_name:
+                self.file_storage_service.update_file(
+                    current_entity.file_name, entity.file, entity.file.content_type
+                )
+            else:
+                entity.file_name = str(uuid4())
```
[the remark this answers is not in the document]
zm2he: Fixed.

backend/python/app/services/implementations/entity_service_mg.py
```diff
@@ -35,7 +45,28 @@ def create_entity(self, entity):
...
+                    file_name, entity.file, entity.file.content_type
+                )
+            else:
+                file_name = str(uuid4())
+                self.file_storage_service.create_file(
+                    file_name, entity.file, entity.file.content_type
+                )
+            entity.__dict__.pop("file", None)
+        elif file_name:
+            self.file_storage_service.delete_file(file_name)
```
sherryhli, line 65: Need to add:
```python
entity.__dict__.pop("file", None)
file_name = None
```

The second line results in a null `file_name` field in MongoDB, would be nice if we can remove the field entirely if there is no file. Not a priority though, feel free to just leave it.

backend/python/app/services/implementations/entity_service.py
```diff
@@ -34,7 +44,31 @@ def create_entity(self, entity):
...
+                    file_name, entity.file, entity.file.content_type
+                )
+            else:
+                file_name = str(uuid4())
+                self.file_storage_service.create_file(
+                    file_name, entity.file, entity.file.content_type
+                )
+            entity.__dict__.pop("file", None)
+        elif file_name:
+            self.file_storage_service.delete_file(file_name)
```
sherryhli, line 66: Need to add:
```python
entity.__dict__.pop("file", None)
file_name = None
```

backend/python/app/rest/entity_routes.py
```diff
@@ -76,7 +88,12 @@ def create_entity():
 # @require_authorization_by_role({"User", "Admin"})
 # def update_entity(id):
 #     try:
-#         body = EntityDTO(**request.json)
+#         if request.content_type == "application/json"
```
sherryhli, line 91:
```suggestion
#         if request.content_type == "application/json":
```

backend/python/app/services/implementations/entity_service_mg.py
tamim-khan: Oh `file_name = None` is a good catch, updated that.

Not passing in `file_name=None` to `Entity.objects(id=id).modify(...)` keeps the old filename. Also tried [this](https://github.com/MongoEngine/mongoengine/issues/2128) but it did not work either.

Also not sure if deleting the field is a better pattern than having it as null because the only way to do it in postgres is to have it as a nullable column.

sherryhli: > Also not sure if deleting the field is a better pattern than having it as null because the only way to do it in postgres is to have it as a nullable column.

Yeah that's fair, let's leave it then 👍

backend/python/app/services/implementations/entity_service.py
```diff
@@ -34,7 +44,32 @@ def create_entity(self, entity):
...
+            file_name = None
```
sherryhli, line 67: This line is needed as well right?

```python
entity.__dict__.pop("file", None)
```

sherryhli (dismissed):
LGTM, thanks for both of your patience in addressing all the changes! I'll take care of merging this

sherryhli (approved):

PR #88, opened 2021-07-14
Integrate TS FileStorageService into EntityService

### LINKED ISSUE #80: Integrate TS FileStorageService into EntityService
**Acceptance Criteria:**

- `Entity` model has been augmented to include a `fileName` and a mapping to a UUID
- `EntityService` CRUD methods call `FileStorageService` CRUD methods
- Update EntityDTOs (EntityRequestDTO is augmented with a file path field, EntityResponseDTO has a new `fileName` field (UUID)

```jsx
export interface EntityResponseDTO {
  id: string;
  stringField: string;
  intField: number;
  enumField: string;
  stringArrayField: string[];
  boolField: boolean;
  fileName: string; // file name (UUID)
}
```

- Update REST and GraphQL endpoints
    - Separate endpoint for file retrieveal (retrieve by looking up with file UUID)
    - Express: look into using `multer`, middleware for accessing multipart formdata, it also configures temp storage (in-memory, and on disk options available). Research if any special considerations need to be taken to send a file requet in GraphQL
    - Flask: follow [https://flask.palletsprojects.com/en/2.0.x/patterns/fileuploads/](https://flask.palletsprojects.com/en/2.0.x/patterns/fileuploads/)
    - Make sure to validate file extenions, only allow 'txt', 'pdf', 'png', 'jpg', 'jpeg', 'gif'
- Annotate with scrubber tags

**Dev Notes:**

Solution:

Rabbit Holes:

alexguo8:
## Notion ticket link

Closes #80 

## Implementation description
* Updated models to have a file name field
* Updated entity request DTO to have a filePath field and response DTO to have a fileName field
* Updated EntityService for Postgres and mongo to call FileStorageService methods
* Updated entity routes , added separate get endpoint for retrieving file by fileUUID
* Added file type validation
* Updated create/update entity GraphQL mutations to take a FileUpload type parameter (documentation https://www.apollographql.com/docs/apollo-server/data/file-uploads/)
* Added separate GraphQL query to retrieve file by fileUUID
* Refactor FileStorageService (changing the bucket field to bucketName field to prevent firebase functions from being called before initializeApp())

## Steps to test
REST:
(using Postman)
1. Set the body to form-data and add the body field (as stringified JSON) and file field (postman lets you upload files).
image

```
{
    "stringField": "test 1",
    "intField": 1,
    "enumField": "A",
    "stringArrayField": ["test1"],
    "boolField": true
}
```
(if you want an example)
2. Send the POST request to entities endpoint to create an entity. Verify in firebase console that the file is created.
3. Try creating another entity whose mime type does not belong to text/plain, application/pdf, image/png, image/jpeg, image/gif. Verify that there is a validation error.
4. Try creating another entity without file parameter. Verify that this works (file field is optional).
5. Send a GET request to /entities endpoint. Verify that the file name is returned.
6. Send a GET request to /entities/files/{fileUUID}. Verify that you get the signed URL.
7. Send a PUT request to this entity's endpoint, with a different file. Verify that the file is updated.
8. Send a PUT request to this entity's endpoint, with no file field. Verify that the file is deleted in firebase console.
9. Send a PUT request to this entity's endpoint, with a file field. Verify that a new file UUID is generated and that it appears in firebase console.
10. Send a DELETE request to this entity's endpoint. Verify that the file is deleted in firebase console.
11. Repeat with the other EntityService (mongo or Postgres)

GraphQL:
1. Download [Altair](https://altair.sirmuel.design) (or use postman with graphql setting, haven't tested it with this though)
2. Add the file at the bottom of the window (click variables)
image
You can use this mutation to create entity.

```
mutation($file: Upload) {
  createEntity(
    entity: {
      stringField: "hi", 
      intField: 2,
      stringArrayField: ["hi"],
      boolField: true,
      enumField: A
    }, file: $file) {
    id
    stringField
    intField
    boolField
    enumField
    stringArrayField
    fileName
  }
}
```
3. You can use this query to get a file signed url.
```
query {
   file(fileUUID: "cat.png")
}
```
4. Repeat similar steps as REST for update entity, you can use this mutation.
```
mutation($file: Upload) {
  updateEntity(
    id: 1,
    entity: {
      stringField: "hi", 
      intField: 2,
      stringArrayField: ["hi"],
      boolField: true,
      enumField: A
    }, file: $file) {
    id
    stringField
    intField
    boolField
    enumField
    stringArrayField
    fileName
  }
}
```
5. Repeat similar step for deleting entity, you can use this mutation
```
mutation {
  deleteEntity(id: 1)
}
```
6. Repeat with the other EntityService (mongo or Postgres)

* Update functionality is complex - does everything look ok?

## Checklist

### DIFF
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -1,7 +1,7 @@
 import fs from "fs";
 import { FileUpload } from "graphql-upload";
 import EntityService from "../../services/implementations/EntityServiceMg";
-import FileStorageService from "../../services/implementations/storageService";
+import FileStorageService from "../../services/implementations/fileStorageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
 
 const defaultBucket = process.env.DEFAULT_BUCKET || "";
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -4,6 +4,10 @@ import { ReadStream } from "fs-capacitor";
 import EntityService from "../../services/implementations/EntityServiceMg";
 import FileStorageService from "../../services/implementations/fileStorageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
+import {
+  validateFileType,
+  getFileTypeValidationError,
+} from "../../middlewares/validators/util";
 
 const defaultBucket = process.env.DEFAULT_BUCKET || "";
 const fileStorageService = new FileStorageService(defaultBucket);
@@ -44,9 +48,12 @@ const entityResolvers = {
         const uploadDir = "uploads";
         filePath = `${uploadDir}/${filename}`;
         fileContentType = mimetype;
+        if (!validateFileType(fileContentType)) {
+          throw new Error(getFileTypeValidationError(fileContentType));
+        }
         await writeFile(createReadStream(), filePath);
       }
-      const newEntity = entityService.createEntity({
+      const newEntity = await entityService.createEntity({
         stringField: entity.stringField,
         intField: entity.intField,
         enumField: entity.enumField,
@@ -75,9 +82,12 @@ const entityResolvers = {
         const uploadDir = "uploads";
         filePath = `${uploadDir}/${filename}`;
         fileContentType = mimetype;
+        if (!validateFileType(fileContentType)) {
+          throw new Error(getFileTypeValidationError(fileContentType));
+        }
         await writeFile(createReadStream(), filePath);
       }
-      const updatedEntity = entityService.updateEntity(id, {
+      const updatedEntity = await entityService.updateEntity(id, {
         stringField: entity.stringField,
         intField: entity.intField,
         enumField: entity.enumField,
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -1,5 +1,5 @@
 import fs from "fs";
-import { FileUpload } from "graphql-upload";
+import { FileUpload, GraphQLUpload } from "graphql-upload";
 import EntityService from "../../services/implementations/EntityServiceMg";
 import FileStorageService from "../../services/implementations/storageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
@@ -69,6 +69,7 @@ const entityResolvers = {
       return entityService.deleteEntity(id);
     },
   },
+  Upload: GraphQLUpload,
 };
 
 export default entityResolvers;
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -16,6 +16,9 @@ const entityResolvers = {
     entities: async () => {
       return entityService.getEntities();
     },
+    file: async (_req: any, { fileUUID }: { fileUUID: string }) => {
+      return fileStorageService.getFile(fileUUID);
+    },
   },
   Mutation: {
     createEntity: async (
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -1,5 +1,6 @@
 import fs from "fs";
 import { FileUpload } from "graphql-upload";
+import { ReadStream } from "fs-capacitor";
 import EntityService from "../../services/implementations/EntityServiceMg";
 import FileStorageService from "../../services/implementations/fileStorageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
@@ -7,6 +8,20 @@ import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
 const defaultBucket = process.env.DEFAULT_BUCKET || "";
 const fileStorageService = new FileStorageService(defaultBucket);
 const entityService = new EntityService(fileStorageService);
+
+const writeFile = (
+  readStream: ReadStream,
+  filePath: string,
+): Promise<void> => {
+  return new Promise((resolve, reject) => {
+    const out = fs.createWriteStream(filePath);
+    readStream.pipe(out);
+    out.on("finish", () => {
+      resolve();
+    });
+    out.on("error", (err) => reject(err));
+  });
+};
 
 const entityResolvers = {
   Query: {
@@ -32,7 +47,7 @@ const entityResolvers = {
         const uploadDir = "uploads";
         filePath = `${uploadDir}/${filename}`;
         fileContentType = mimetype;
-        createReadStream().pipe(fs.createWriteStream(filePath));
+        await writeFile(createReadStream(), filePath);
       }
       return entityService.createEntity({
         stringField: entity.stringField,
@@ -59,7 +74,7 @@ const entityResolvers = {
         const uploadDir = "uploads";
         filePath = `${uploadDir}/${filename}`;
         fileContentType = mimetype;
-        createReadStream().pipe(fs.createWriteStream(filePath));
+        await writeFile(createReadStream(), filePath);
       }
       return entityService.updateEntity(id, {
         stringField: entity.stringField,
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -1,7 +1,12 @@
+import fs from "fs";
+import { FileUpload } from "graphql-upload";
 import EntityService from "../../services/implementations/EntityServiceMg";
+import FileStorageService from "../../services/implementations/storageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
 
-const entityService = new EntityService();
+const defaultBucket = process.env.DEFAULT_BUCKET || "";
+const fileStorageService = new FileStorageService(defaultBucket);
+const entityService = new EntityService(fileStorageService);
 
 const entityResolvers = {
   Query: {
@@ -15,26 +20,46 @@ const entityResolvers = {
   Mutation: {
     createEntity: async (
       _req: any,
-      { entity }: { entity: EntityRequestDTO },
+      { entity, file }: { entity: EntityRequestDTO; file: FileUpload },
     ) => {
+      let filePath = "";
+      if (file) {
+        const { createReadStream } = file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${file.filename}`;
+        createReadStream().pipe(fs.createWriteStream(filePath));
+      }
       return entityService.createEntity({
         stringField: entity.stringField,
         intField: entity.intField,
         enumField: entity.enumField,
         stringArrayField: entity.stringArrayField,
         boolField: entity.boolField,
+        filePath,
       });
     },
     updateEntity: async (
       _req: any,
-      { id, entity }: { id: string; entity: EntityRequestDTO },
+      {
+        id,
+        entity,
+        file,
+      }: { id: string; entity: EntityRequestDTO; file: FileUpload },
     ) => {
+      let filePath = "";
+      if (file) {
+        const { createReadStream } = file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${file.filename}`;
+        createReadStream().pipe(fs.createWriteStream(filePath));
+      }
       return entityService.updateEntity(id, {
         stringField: entity.stringField,
         intField: entity.intField,
         enumField: entity.enumField,
         stringArrayField: entity.stringArrayField,
         boolField: entity.boolField,
+        filePath,
       });
     },
     deleteEntity: async (_req: any, { id }: { id: string }) => {
--- a/backend/typescript/graphql/resolvers/entityResolvers.ts
+++ b/backend/typescript/graphql/resolvers/entityResolvers.ts
@@ -1,5 +1,5 @@
 import fs from "fs";
-import { FileUpload, GraphQLUpload } from "graphql-upload";
+import { FileUpload } from "graphql-upload";
 import EntityService from "../../services/implementations/EntityServiceMg";
 import FileStorageService from "../../services/implementations/storageService";
 import { EntityRequestDTO } from "../../services/interfaces/IEntityService";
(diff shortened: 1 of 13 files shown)


backend/typescript/package.json
```diff
@@ -61,5 +67,9 @@
     "nodemon": "^2.0.7",
     "prettier": "^2.2.1",
     "typescript": "^4.1.5"
+  },
+  "resolutions": {
```
alexguo8, line 71: This is because of an issue with apollo-server-express using an older version of graphql-upload - read more here https://github.com/apollographql/apollo-server/issues/3508

backend/typescript/models/entity.pgmodel.ts
```diff
@@ -18,4 +18,7 @@ export default class Entity extends Model {
 
   @Column
   bool_field!: boolean;
+
+  @Column
+  file_name!: string;
```
alexguo8, line 23: The reason why this has ! and mongo model does not have required = true is because the field in mongo can't be set to "" if it's required, whereas in Postgres it can

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -106,18 +154,21 @@ class EntityService implements IEntityService {
...
+      const entityToDelete = await PgEntity.findByPk(id, { raw: true });
+      const deleteResult: number | null = await PgEntity.destroy({
         where: { id },
       });
 
-      if (!deletedEntity) {
+      if (!entityToDelete || !deleteResult) {
         throw new Error(`Entity id ${id} not found`);
       }
+      this.storageService.deleteFile(entityToDelete.file_name);
```
tamim-khan, line 171: We would get a lingering file if this last line fails

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -51,14 +62,23 @@ class EntityService implements IEntityService {
...
         bool_field: entity.boolField,
+        file_name: fileName,
       });
+      if (entity.filePath) {
+        this.storageService.createFile(
+          fileName,
+          entity.filePath,
+          entity.fileContentType,
+        );
+      }
```
tamim-khan, line 81: Similar issue here where the file may not be created but the entity is created

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -95,6 +126,23 @@ class EntityService implements IEntityService {
...
+        } else {
+          this.storageService.createFile(
+            fileName,
+            entity.filePath,
+            entity.fileContentType,
+          );
+        }
+      } else if (currentFileName) {
+        this.storageService.deleteFile(currentFileName);
+      }
```
tamim-khan, line 145: Similar issue here where the file may not be updated but the entity is updated

tamim-khan (commented):
One way of fixing this issue is by having the create, update and delete in transactions, so they can be rolled back if the file storage service is down

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -100,6 +148,7 @@ class EntityService implements IEntityService {
       if (!deletedEntity) {
         throw new Error(`Entity id ${id} not found`);
       }
+      this.storageService.deleteFile(deletedEntity.fileName);
```
sherryhli, line 151: This throws an error if `deletedEntity` doesn't have an associated file to begin with right? Don't think it should be an error condition, let it silently fail

backend/typescript/services/implementations/EntityServicePg.ts
sherryhli, line 81: Let's do the file creation before the entity creation like in the Mongo implementation. It's possible to get a dangling file if entity creation fails, but I think that's more preferable than having a dangling file reference (i.e. file creation fails **after** entity creation succeeds, entity's fileUUID does not point to any file)

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -72,14 +92,41 @@ class EntityService implements IEntityService {
...
+        } else {
+          this.storageService.createFile(
+            fileName,
+            entity.filePath,
+            entity.fileContentType,
+          );
+        }
+      } else if (currentFileName) {
+        this.storageService.deleteFile(currentFileName);
+      }
```
sherryhli, line 129: Let's do file update/creation before entity update. Potential dangling file issue if we hit the `createFile` case: creation could fail and there would be a reference to a file that does not exist in the DB

Same thing for the PG service

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -106,18 +154,21 @@ class EntityService implements IEntityService {
...
   async deleteEntity(id: string): Promise<void> {
     try {
-      const deletedEntity: number | null = await PgEntity.destroy({
+      const entityToDelete = await PgEntity.findByPk(id, { raw: true });
+      const deleteResult: number | null = await PgEntity.destroy({
         where: { id },
       });
 
-      if (!deletedEntity) {
+      if (!entityToDelete || !deleteResult) {
```
sherryhli, line 168: Isn't the `!deleteResult` check sufficient?

sherryhli (changes requested):
Awesome work! Did a first pass, only significant thing to address is the dangling file issue

backend/typescript/services/implementations/EntityServicePg.ts
alexguo8: In practice it should be, I just added it in because entityToDelete has type PgEntity | null, so I wanted to remove the TS error in the deleteFile call
Is there a better way to handle this?

backend/typescript/services/implementations/EntityServiceMg.ts
alexguo8: Do you mean I should update the FileStorageService or add another try catch in this function for that case?

sherryhli: I think either another try-catch, or wrapping it in `if (deletedEntity.fileName) { ... }` are fine

backend/typescript/services/implementations/EntityServicePg.ts
sherryhli: Ok I see, this works 👍

backend/typescript/middlewares/validators/entityValidators.ts
```diff
@@ -11,22 +13,31 @@ export const entityRequestDtoValidator = async (
   res: Response,
   next: NextFunction,
 ) => {
-  if (!validatePrimitive(req.body.stringField, "string")) {
+  let body;
+  try {
+    body = JSON.parse(req.body.body);
+  } catch (e) {
+    return res.status(500).send(e.message);
```
sherryhli, line 20: `400` status code (bad request) is probably more suitable

backend/typescript/middlewares/validators/util.ts
```diff
@@ -1,5 +1,13 @@
 type Type = "string" | "integer" | "boolean";
 
+const allowableContentTypes = [
```
sherryhli, line 3: Nit: this can be a set

backend/typescript/services/implementations/EntityServiceMg.ts
sherryhli: Actually let's just do the if statement. If the `filename === ""`, that means there isn't an associated file and we should not even attempt to delete and log an error. If there is a `filename` but it doesn't exist in storage for some reason, that's an error case (this case is already accounted for here).

sherryhli (commented):
Second pass done, one more round and I think we'll be good 🚀

backend/typescript/graphql/resolvers/entityResolvers.ts
```diff
@@ -11,30 +16,59 @@ const entityResolvers = {
...
+      { entity, file }: { entity: EntityRequestDTO; file: Promise<FileUpload> },
     ) => {
+      let filePath = "";
+      let fileContentType = "";
+      if (file) {
+        const { createReadStream, mimetype, filename } = await file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${filename}`;
+        fileContentType = mimetype;
+        createReadStream().pipe(fs.createWriteStream(filePath));
```
sherryhli, line 35: Should probably await the completion of the write like in the docs: https://www.apollographql.com/docs/apollo-server/data/file-uploads/

```ts
import { finished } = from "stream/promises";

const out = fs.createWriteStream(filePath);
createReadStream().pipe(out);
await finished(out);
```

backend/typescript/graphql/resolvers/entityResolvers.ts
```diff
@@ -11,30 +16,59 @@ const entityResolvers = {
...
+      }
```
sherryhli, line 36: Could delete the temp file in `uploadDir` before returning right?

backend/typescript/graphql/resolvers/entityResolvers.ts
```diff
@@ -11,31 +28,68 @@ const entityResolvers = {
...
+      let filePath = "";
+      let fileContentType = "";
+      if (file) {
+        const { createReadStream, mimetype, filename } = await file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${filename}`;
+        fileContentType = mimetype;
+        await writeFile(createReadStream(), filePath);
+      }
+      const newEntity = entityService.createEntity({
```
sherryhli, line 49: Must await if not returning result directly
```suggestion
      const newEntity = await entityService.createEntity({
```

backend/typescript/graphql/resolvers/entityResolvers.ts
```diff
@@ -11,31 +28,68 @@ const entityResolvers = {
...
+      let filePath = "";
+      let fileContentType = "";
+      if (file) {
+        const { createReadStream, mimetype, filename } = await file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${filename}`;
+        fileContentType = mimetype;
+        await writeFile(createReadStream(), filePath);
+      }
+      const updatedEntity = entityService.updateEntity(id, {
```
sherryhli, line 80:
```suggestion
      const updatedEntity = await entityService.updateEntity(id, {
```

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -51,8 +62,16 @@ class EntityService implements IEntityService {
 
   async createEntity(entity: EntityRequestDTO): Promise<EntityResponseDTO> {
     let newEntity: Entity | null;
+    const fileName = entity.filePath ? uuidv4() : "";
     try {
-      newEntity = await MgEntity.create(entity);
+      if (entity.filePath) {
+        this.storageService.createFile(
```
sherryhli, line 68:
```suggestion
        await this.storageService.createFile(
```

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -72,11 +92,36 @@ class EntityService implements IEntityService {
...
-      updatedEntity = await MgEntity.findByIdAndUpdate(id, entity, {
-        new: true,
-        runValidators: true,
-      });
+      const currentEntity = await MgEntity.findById(id, "fileName");
+      const currentFileName = currentEntity?.fileName;
+      if (entity.filePath) {
+        fileName = currentFileName || uuidv4();
+        if (currentFileName) {
+          this.storageService.updateFile(
```
sherryhli, line 102:
```suggestion
          await this.storageService.updateFile(
```

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -72,11 +92,36 @@ class EntityService implements IEntityService {
...
+            fileName,
+            entity.filePath,
+            entity.fileContentType,
+          );
+        } else {
+          this.storageService.createFile(
```
sherryhli, line 108:
```suggestion
          await this.storageService.createFile(
```

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -72,11 +92,36 @@ class EntityService implements IEntityService {
...
+            fileName,
+            entity.filePath,
+            entity.fileContentType,
+          );
+        }
+      } else if (currentFileName) {
+        this.storageService.deleteFile(currentFileName);
```
sherryhli, line 115:
```suggestion
        await this.storageService.deleteFile(currentFileName);
```

backend/typescript/services/implementations/EntityServiceMg.ts
```diff
@@ -100,6 +146,9 @@ class EntityService implements IEntityService {
       if (!deletedEntity) {
         throw new Error(`Entity id ${id} not found`);
       }
+      if (deletedEntity.fileName) {
+        this.storageService.deleteFile(deletedEntity.fileName);
```
sherryhli, line 150:
```suggestion
        await this.storageService.deleteFile(deletedEntity.fileName);
```

backend/typescript/services/implementations/EntityServicePg.ts
```diff
@@ -51,13 +62,22 @@ class EntityService implements IEntityService {
 
   async createEntity(entity: EntityRequestDTO): Promise<EntityResponseDTO> {
     let newEntity: PgEntity | null;
+    const fileName = entity.filePath ? uuidv4() : "";
     try {
+      if (entity.filePath) {
+        this.storageService.createFile(
```
sherryhli, line 68: Same thing as the mongo service, please add await before all calls to `storageService` methods :)

backend/typescript/graphql/resolvers/entityResolvers.ts
```diff
@@ -11,31 +28,68 @@ const entityResolvers = {
...
+      { entity, file }: { entity: EntityRequestDTO; file: Promise<FileUpload> },
     ) => {
-      return entityService.createEntity({
+      let filePath = "";
+      let fileContentType = "";
+      if (file) {
+        const { createReadStream, mimetype, filename } = await file;
+        const uploadDir = "uploads";
+        filePath = `${uploadDir}/${filename}`;
+        fileContentType = mimetype;
```
sherryhli, line 46: Let's validate the the content-type, can reuse the `validateFileType` function used by REST middleware. Same thing for the update mutation.

sherryhli (commented):
Almost there! Tested and it looks good for the most part, just need to remember to await all calls to `FileStorageService` methods since they're all async. Easy to miss until the temp file cleanup logic was added (file gets deleted before it gets written to cloud storage) :')

sherryhli (approved):
Awesome! There's a lot in here, thanks for patiently addressing all the changes 🚢

Issue #65, opened 2021-07-07
Create backend utility to generate csv from tabular data, add option to return csv from bulk GET endpoints


DinuWije:
## Context:

[Research, compare & contrast CSV generation from frontend and backend](https://www.notion.so/Research-compare-contrast-CSV-generation-from-frontend-and-backend-820a572e1c9447ffbdef97a2a4731120)

**Why backend approach?**

- The backend vs frontend decision isn't actually very clear-cut, difficult to generalize all csv export use cases
- Backend implementation keeps logic in the frontend lightweight, but might mean making an unnecessary request to retrieve csv data when JSON may already be available in the frontend. This approach is best-suited for when the frontend does not require the JSON representation of the csv data at all

## Acceptance Criteria

- GET /entities and GET /users accept `"text/csv"` content type and will return data in csv form
- New GraphQL queries created for entities and users, which returns a csv string
- There is a utility function in the `/utilities` folders that will take arbitrary tabular data (i.e. array of objects or dicts) as an argument and return the data in csv form

```tsx
generateCSV(data: Array<{[key: string]: any}>): string
```

## Dev Notes

- To implement the `generateCSV` function, consider using a third-party library (see research notes for suggestions) instead of building from scratch → possibly easier handling of nested objects, arrays
- REST endpoints: retrieve data from services as usual, but if `req.headers.content-type` is `text/csv`, pass the data to `generateCSV` and return that result instead
- GraphQL: the new csv queries will retrieve data as usual from the service methods, but will pass the data to `generateCSV` and return that result

## Rabbit Holes:

- Consider nested object representations, max level of nesting

PR #91, opened 2021-07-25
Add backend TS csv generation utility

emilyslouie:
## Notion ticket link

[TS Backend CSV generation utility](https://github.com/uwblueprint/starter-code-v2/issues/65)

## Implementation description
* Wrapper on the json2csv function similar to the frontend
* A response util function that either looks for an `application/json` or `text/csv` content-type to determine what type of response to send back

## Steps to test
REST (using Postman)
1. Set the content-type in the headers to `text/csv` and the request type to GET
2. Test `http://localhost:5000/entities` and `http://localhost:5000/users` to ensure you get back number of items in CSV format

Graphql (using Altair)
1. Set the request type to GET and put in this URL `http://localhost:5000/graphql`
2. Run the following queries and ensure you get the same CSV string as the REST responses

```
query {
   usersCSV
} // gets all users

query {
   entitiesCSV
} // should return all entities
```

* Might have missed this but do we need options similar to the frontend PR?

## Checklist


backend/typescript/rest/entityRoutes.ts
```diff
@@ -49,23 +50,32 @@ entityRouter.post(
...
+    await setResponseByMimeType(res, 500, contentType, [
+      {
+        error: e.message,
+      },
+    ]);
   }
 });
 
 /* Get entity by id */
 entityRouter.get("/:id", async (req, res) => {
```
sherryhli, line 67: Don't think we need to allow `text/csv` requests for GET requests for singular entities :)

backend/typescript/rest/userRoutes.ts
```diff
@@ -16,49 +17,70 @@ const userService: IUserService = new UserService();
...
     }
     return;
   }
 
   if (userId) {
     if (typeof userId !== "string") {
-      res
-        .status(400)
-        .json({ error: "userId query parameter must be a string." });
+      await setResponseByMimeType(res, 400, contentType, [
```
sherryhli, line 47: When a `userId` query param is provided we only return a single user so the CSV option is not needed. Sorry for the extra work!

backend/typescript/rest/userRoutes.ts
```diff
@@ -16,49 +17,70 @@ const userService: IUserService = new UserService();
...
     }
     return;
   }
 
   if (email) {
     if (typeof email !== "string") {
-      res
-        .status(400)
-        .json({ error: "email query parameter must be a string." });
+      await setResponseByMimeType(res, 400, contentType, [
```
sherryhli, line 69: Like above, we don't need to support a CSV option

backend/typescript/graphql/types/userType.ts
```diff
@@ -31,8 +31,11 @@ const userType = gql`
 
   extend type Query {
     userById(id: ID!): UserDTO!
+    userByIdCSV(id: ID!): String!
     userByEmail(email: String!): UserDTO!
+    userByEmailCSV(email: String!): String!
```
sherryhli, line 36: Sorry again for the extra work and being unclear in the ticket, we only need the `usersCSV` mutation for bulk user requests.

backend/typescript/graphql/types/entityType.ts
```diff
@@ -32,7 +32,9 @@ const entityType = gql`
 
   extend type Query {
     entity(id: ID!): EntityResponseDTO!
+    entityCSV(id: ID!): String!
```
sherryhli, line 35: No needed

sherryhli (changes requested):
Looks great, love the response util to avoid repeated if-elses! Just a couple of really minor things to address :)

sherryhli (approved):
LGTM! Please hold off on merging though as explained here: https://uwblueprint.slack.com/archives/C01JSU8JJCB/p1628441736003000

PR #104, opened 2021-08-11
Create Typescript registration flow with email verification

alexguo8:
## Notion ticket link

Closes #92 

## Implementation description
* sendEmailVerificationLink function added to auth service instead, to make it more consistent with the reset password function, and so we don’t need the user service to have its own email service instance
* Downside is that in userRoutes, we have to instantiate an authService
* Note: existing admin users may have to send themselves a verification email link so that they can access endpoints once this is merged
* I did this just by adding `await authService.sendEmailVerificationLink("your email");` to the login endpoint and querying once

## Steps to test
# GraphQL
1. Start the server and navigate to http://localhost:5000/graphql . Use this mutation to register a user and verify that the user is created successfully in MongoDB atlas/postgres and in firebase console. Also verify that you are sent a verification email.
```
mutation Register {
  register(user: {
    firstName: "Test",
    lastName: "User",
    email: "insertpersonalemail",
    password: "insertpassword"
  }) {
    firstName
    id
    lastName
    email
    role
    accessToken
  }
}
```
image
2. Try using the access token returned to query entities. Verify that you are unauthorized (since you have not verified email).

```
query getentities {
  entities {
    id
  }
}
```

image
3. Verify your email by clicking the link. 
image

4. Try running the entities query again. Verify that it still fails (because this access token was sent before verification, so the token still thinks the user is unverified)
5. Try logging in and using the new access token to get entities. Verify that it succeeds.

```
mutation login {
  login(email: "insertpersonalemail", password: "insertpassword") {
    accessToken
  }
}
```

6. Create a new user using the following mutation and verify that an email verification link is sent.

```
mutation createUser {
  createUser(user: {
    firstName: "Test",
    lastName: "User",
    email: "insertemail",
    password: "insertpassword",
    role: Admin
  }) {
    firstName
    id
    lastName
    email
    role
  }
}
```

# REST
1. Post the following JSON body to localhost:5000/auth/register, and verify that the user is created in MongoDB atlas/postgres and in firebase console. Also verify that you receive an email verification link.
```
{
	"firstName": "Test",
	"lastName": "User",
	"email": "insertemail",
	"password": "insertpassword"
}
```
2. Try sending a GET request to localhost:5000/entities , using the returned access token. Verify that it fails.
3. Verify your email by clicking the link.
4. Try sending a GET request to localhost:5000/entities with the same token and verify that it fails.
5. Login and use the new token to get entities, and verify that it works.
[...]

PR #105, opened 2021-08-12
Create Python registration flow with email verification

### LINKED ISSUE #93: Python Backend: Create user sign-up (registration) flow with email verification
## Context

Currently, starter code only allows authenticated users with `Admin` roles to create user accounts. We want a separate sign-up (registration) flow which allows any unauthenticated user to create an account with `User` role and then be automatically authenticated.

Additionally, any created account should be sent a email verification link. Until that email address is verified, the user's access token cannot be used to access any application resources.

## Acceptance Criteria
* There is a new `/auth/register` REST endpoint. It is very similar to POST `/user`, but will not require an access token, will only allow creating the `User` role, and will additionally return an `AuthenticatedUser` object containing the access and refresh tokens
* `UserService.create_user` is modified so that Firebase account creation will trigger an email verification link to be sent
* The auth middleware is modified so that only verified users are considered to be authenticated and authorized

## Dev Notes

* Planet Read PR with a subset of these changes: https://github.com/uwblueprint/planet-read/pull/7
* Generate email verification link with Firebase: https://firebase.google.com/docs/auth/admin/email-action-links
* Look at the password reset method in `AuthService` to see how to use `EmailService` to send emails

emilyslouie:
## Notion ticket link

Closes #93 

## Implementation description
* `sendEmailVerificationLink` function added to the auth service
* similar to #104 
* created a `RegisterUserDTO` in resources to validate requests

## Steps to test
1. Post the following JSON body to localhost:5000/auth/register, and verify that the user is created in MongoDB atlas/postgres and in firebase console. Also verify that you receive an email verification link.
```
{
	"first_name": "Test",
	"last_name": "User",
	"email": "insertemail",
	"password": "insertpassword"
}
```
2. Try sending a GET request to localhost:5000/entities , using the returned access token. Verify that it fails.
3. Verify your email by clicking the link.
4. Try sending a GET request to localhost:5000/entities with the same token and verify that it fails.
5. Login and use the new token to get entities, and verify that it works.
6. Try sending a POST request to localhost:5000/users and verify that a user is created, and that a verification email is sent.
```
{
	"first_name": "Test",
	"last_name": "User",
	"email": "insertemail",
	"password": "insertpassword",
        "role": "Admin"
}
```

* 

## Checklist

### DIFF
--- a/backend/python/app/middlewares/validate.py
+++ b/backend/python/app/middlewares/validate.py
@@ -3,11 +3,13 @@ from functools import wraps
 
 from ..resources.create_user_dto import CreateUserDTO
 from ..resources.entity_dto import EntityDTO
+from ..resources.register_user_dto import RegisterUserDTO
 from ..resources.update_user_dto import UpdateUserDTO
 
 dtos = {
     "CreateUserDTO": CreateUserDTO,
     "EntityDTO": EntityDTO,
+    "RegisterUserDTO": RegisterUserDTO,
     "UpdateUserDTO": UpdateUserDTO,
 }
 
--- a/backend/python/app/resources/register_user_dto.py
+++ b/backend/python/app/resources/register_user_dto.py
@@ -1,0 +1,18 @@
+class RegisterUserDTO:
+    def __init__(self, **kwargs):
+        self.first_name = kwargs.get("first_name")
+        self.last_name = kwargs.get("last_name")
+        self.email = kwargs.get("email")
+        self.password = kwargs.get("password")
+
+    def validate(self):
+        error_list = []
+        if type(self.first_name) is not str:
+            error_list.append("The first_name supplied is not a string.")
+        if type(self.last_name) is not str:
+            error_list.append("The last_name supplied is not a string.")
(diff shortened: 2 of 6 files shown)


sherryhli (dismissed):
Looks great! I'll take care of merging this

Copying over a comment I made on Alex's PR, no action needed, just an FYI :)

> Try running the entities query again. Verify that it still fails (because this access token was sent before verification, so the token still thinks the user is unverified)

This part is a little weird, looks like the `email_verified` flag is encoded in the token (can check with https://jwt.io). Ideally the user does not need to authenticate again after verification, otherwise this defeats the purpose of returning a token from the register endpoint. For the time being, the frontend can ignore the token from the register response and just prompt the user to verify and login. Let's leave the token generation code in though in case we can find a workaround.

sherryhli:
Update: we discussed a workaround during today's work session, let's make this change :) See https://github.com/uwblueprint/starter-code-v2/pull/104#issuecomment-899083804

sherryhli (approved):

PR #114, opened 2021-08-26
Integrate Google OAuth into TS

### LINKED ISSUE #102: TS backend and frontend: integrate Google OAuth sign-in
## Context

We want to allow sign-ins with Google OAuth credentials as this is a common ask for Blueprint projects. With Google OAuth, Google is the identity provider, meaning it is responsible for authenticating users.

## Acceptance Criteria
* Users should be able to sign-up and sign-in to the application using a Google Account
* This sign-in flow is implemented (do a check to see if account already exists in Firebase):
    * The frontend authenticates a user with their Google account by directly making an auth request to Google (look into the `react-google-login` library). This request should return a Google `idToken` would identifies the user.
    * Authenticate with our own backend by sending the `idToken`
    * Backend uses [this REST endpoint](https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-with-oauth-credential) to generate access + refresh tokens (from this point onwards the auth flow is exactly the same as the current implementation)
    * Returns token pair to frontend
* Sign-up flow: similar to sign-in but this code path is followed if the user does not yet have a Firebase account
    * Everything above, but include a step for creating the user (both in Firebase and MongoDB/PostgreSQL) before generating the access + refresh token pair
* Implement both flows with the REST API and GraphQL API

## Dev Notes

* Planet Read PR that adds this exact feature, albeit using Python + GraphQL: https://github.com/uwblueprint/planet-read/pull/47 -> the general flow should be very similar
* Firebase REST API reference for the sign-in with OAuth credential endpoint: https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-with-oauth-credential. Note we need to use the REST API because the server-side SDK does not provide this functionality. Add a wrapper over this endpoint in `FirebaseRestClient`.

tamim-khan:
## Ticket link
Closes #102 

## Implementation description
* Mirrors #112

## Steps to test
1. Same as Python PR

* 

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/authService.ts
+++ b/backend/typescript/services/implementations/authService.ts
@@ -44,14 +44,15 @@ class AuthService implements IAuthService {
         idToken,
       );
       const token = {
-        accessToken: idToken,
+        accessToken: googleUser.idToken,
         refreshToken: googleUser.refreshToken,
       };
-      let user = await this.userService.getUserByEmail(googleUser.email);
-      if (user) {
+      try {
+        const user = await this.userService.getUserByEmail(googleUser.email);
         return { ...token, ...user };
-      }
-      user = await this.userService.createUser(
+      } catch (error) {}
+
+      const user = await this.userService.createUser(
         {
           firstName: googleUser.firstName,
           lastName: googleUser.lastName,
--- a/backend/typescript/services/implementations/authService.ts
+++ b/backend/typescript/services/implementations/authService.ts
@@ -37,6 +37,39 @@ class AuthService implements IAuthService {
     }
   }
 
+  /* eslint-disable class-methods-use-this */
+  async generateTokenOAuth(idToken: string): Promise<AuthDTO> {
+    try {
+      const googleUser = await FirebaseRestClient.signInWithGoogleOAuth(
+        idToken,
+      );
+      const token = {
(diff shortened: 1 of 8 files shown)


backend/typescript/services/implementations/authService.ts
```diff
@@ -37,6 +37,39 @@ class AuthService implements IAuthService {
...
   }
 
+  /* eslint-disable class-methods-use-this */
+  async generateTokenOAuth(idToken: string): Promise<AuthDTO> {
+    try {
+      const googleUser = await FirebaseRestClient.signInWithGoogleOAuth(
+        idToken,
+      );
+      const token = {
+        accessToken: idToken,
```
alexguo8, line 47: Should be googleUser.idToken

backend/typescript/services/implementations/authService.ts
```diff
@@ -37,6 +37,39 @@ class AuthService implements IAuthService {
...
+        refreshToken: googleUser.refreshToken,
+      };
+      let user = await this.userService.getUserByEmail(googleUser.email);
```
alexguo8, line 50: Should be wrapped in try catch so that we can create user if not found

alexguo8 (approved):

PR #133, opened 2021-11-01
Fix error when deleting an entity with no file in Python

alexguo8:
## Implementation description
* When deleting an entity with no file in Python, there is an error saying `Failed to delete file None. Reason = None could not be converted to unicode`
* Updated response code for delete entity to be consistent with TypeScript implementation

## Steps to test
1. Use Python and delete an entity that has no file

* I remember discussing the response code for delete before but I can't remember why we use 200 instead of 204, was there a reason?

## Checklist

### DIFF
--- a/backend/typescript/rest/entityRoutes.ts
+++ b/backend/typescript/rest/entityRoutes.ts
@@ -116,7 +116,7 @@ entityRouter.delete("/:id", async (req, res) => {
 
   try {
     await entityService.deleteEntity(id);
-    res.status(204).send();
+    res.status(200).send();
   } catch (e) {
     res.status(500).send(e.message);
   }
--- a/backend/typescript/rest/entityRoutes.ts
+++ b/backend/typescript/rest/entityRoutes.ts
@@ -115,8 +115,8 @@ entityRouter.delete("/:id", async (req, res) => {
   const { id } = req.params;
 
   try {
-    await entityService.deleteEntity(id);
-    res.status(200).send();
+    const deletedId = await entityService.deleteEntity(id);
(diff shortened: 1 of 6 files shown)


backend/python/app/rest/entity_routes.py
```diff
@@ -169,7 +169,7 @@ def delete_entity(id):
         error_message = getattr(e, "message", None)
         return jsonify({"error": (error_message if error_message else str(e))}), 500
 
-    return jsonify(result), 200
+    return jsonify(result), 204
```
sherryhli, line 172: 200 makes more sense because we're returning content here (ID of the deleted entity)

> I remember discussing the response code for delete before but I can't remember why we use 200 instead of 204, was there a reason?

I think we actually want to converge on always returning 200. GraphQL best practice is to always return a value from queries/mutations so for parity, REST should also return a value (only some(?) of our GraphQL delete mutations are currently returning values, but it's something we should fix 😅)

backend/typescript/rest/entityRoutes.ts
```diff
@@ -116,7 +116,7 @@ entityRouter.delete("/:id", async (req, res) => {
 
   try {
     await entityService.deleteEntity(id);
-    res.status(204).send();
+    res.status(200).send();
```
sherryhli, line 119: We can return the deleted id now right?

```suggestion
    const deletedId = await entityService.deleteEntity(id);
    res.status(200).json({ id: deletedId });
```

sherryhli (dismissed):
Sorry, I forgot about this PR. Just a minor suggestion!

sherryhli (approved):
🚢

PR #136, opened 2021-12-18
Return string id from TypeScript PostgreSQL user and entity services

sherryhli:
## Notion ticket link

N/A

## Implementation description
Cast type of `id` to string in responses from the PostgreSQL user and entity services (the `id` is typed as `string` in the user and entity DTOs but we somehow got past typechecking before and returned numbers).

## Steps to test
1. Modify `backend/typescript/rest/userRoutes.ts`, `backend/typescript/rest/authRoutes.ts`, and `backend/typescript/middlewares/auth.ts` to import the PostgreSQL version of `UserService`
2. Comment out the `py-backend` service in `docker-compose.yml`
3. Run the application
```bash
docker-compose up --build
```
4. Test all `/users` and `/entities` endpoints and verify that the `id`s in responses are strings

Instead of this approach of always converting the `id` to string, should we define separate DTOs for PostgreSQL services where the `id` is number? My justification for current approach:
* More flexible if teams choose to use UUIDs instead of ints for the primary key
* In line with usage by GraphQL resolvers, which automatically converts the `id` in the DTO to be a string in the response because we mapped it to the GraphQL type `ID`
* Less work as we can share the exact same DTO typedef with the MongoDB implementation 😅

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/EntityServicePg.ts
+++ b/backend/typescript/services/implementations/EntityServicePg.ts
@@ -32,7 +32,7 @@ class EntityService implements IEntityService {
     }
 
     return {
-      id: entity.id,
+      id: String(entity.id),
       stringField: entity.string_field,
       intField: entity.int_field,
       enumField: entity.enum_field,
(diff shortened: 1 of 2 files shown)


alexguo8 (approved):
> Instead of this approach of always converting the id to string, should we define separate DTOs for PostgreSQL services where the id is number? My justification for current approach:
More flexible if teams choose to use UUIDs instead of ints for the primary key
In line with usage by GraphQL resolvers, which automatically converts the id in the DTO to be a string in the response because we mapped it to the GraphQL type ID
Less work as we can share the exact same DTO typedef with the MongoDB implementation 😅

I agree with these points, LGTM!

PR #168, opened 2022-03-26
Add Simple Entity TypeScript API and frontend

alexguo8:
## Notion ticket link
Apologies for the diff, majority of it is just copied from existing entity files

## Implementation description
* Created copies of existing entity API and frontend, without file storage

## Steps to test
1. Use the scrubber locally (yarn dev -t in starter-code-cli) with a TypeScript backend (and any API or database).
2. Login on http://localhost:3000 and verify that the new Create Simple Entity, Update Simple Entity, and Display Simple Entities pages work as expected.
3. Verify that the existing entity API is unaffected.

## What should reviewers focus on?
*

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #169, opened 2022-04-02
Add Simple Entity Python API

alexguo8:
## Notion ticket link

N/A

## Implementation description
* Created copies of existing entity API and frontend, without file storage
* Python version of this PR https://github.com/uwblueprint/starter-code-v2/pull/168

## Steps to test
1. Use the scrubber locally (yarn dev -t in starter-code-cli) with a Python backend (and any database).
2. Login on http://localhost:3000/ and verify that the new Create Simple Entity, Update Simple Entity, and Display Simple Entities pages work as expected.
3. Verify that the existing entity API is unaffected.

* 

## Checklist


backend/python/app/models/simple_entity.py
```diff
@@ -0,0 +1,71 @@
...
+
+# common columns and methods across multiple data models can be added via a Mixin class:
+# https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/mixins.html
+
+# see examples of Mixins in current and past Blueprint projects:
+# https://github.com/uwblueprint/dancefest-web/blob/master/db/models.py#L10-L70
+# https://github.com/uwblueprint/plasta/blob/master/backend/app/models/mixins.py#L10-L95
+
+
+class SimpleEntity(db.Model):
```
sherryhli, line 16: Do you mind creating a migration for this table as well?
https://uwblueprint.github.io/starter-code-v2/docs/migrations/python-migrations.html#how-to-create-and-run-a-migration

backend/python/app/rest/simple_entity_routes.py
```diff
@@ -0,0 +1,167 @@
...
+    "header": True,
+    "flatten_lists": False,
+    "flatten_objects": False,
+}
+
+# define instance of SimpleEntityService
+simple_entity_service = SimpleEntityService(current_app.logger)
+
+# defines a shared URL prefix for all routes
+blueprint = Blueprint("simple_entity", __name__, url_prefix="/simpleEntities")
```
sherryhli, line 21: Nit
```suggestion
blueprint = Blueprint("simple_entity", __name__, url_prefix="/simple-entities")
```

backend/python/app/rest/__init__.py
```diff
@@ -1,14 +1,14 @@
 def init_app(app):
-    # auth {
-    from . import user_routes, auth_routes, entity_routes, documentation_routes
-    # } auth
-    # no-auth {
-    from . import entity_routes, documentation_routes
-    # } no-auth
+    from . import (
```
sherryhli, line 2: Does this PR include removing **all** auth and file-storage tags from the Python backend? Wondering if this conflicts with what @matthewgeng is working on

backend/python/app/middlewares/validate.py
```diff
@@ -50,10 +51,6 @@ def wrapper(*args, **kwargs):
                 req = json.loads(req_body)
                 req["file"] = request.files.get("file", default=None)
                 dto = dtos[dto_class_name](**req)
-            # } file-storage
-            # no-file-storage {
-            dto = dtos[dto_class_name](**request.json)
```
sherryhli, line 55: I think this means that requests to the simple-entities endpoint must specify `application/json` as the content type (i.e. a value must be supplied, can't rely on default) in order to follow the correct code path for validation. Could you please add a comment about this in the routes file and the Swagger docs?

sherryhli (changes requested):
Just a couple of small things :)

backend/python/app/rest/__init__.py
alexguo8: No, only files I had to modify

backend/python/app/middlewares/validate.py
alexguo8: Axios automatically sets the Content-Type header to `application/json` when the request is a JS object
https://masteringjs.io/tutorials/axios/post-json

backend/python/app/models/enum.py
```diff
@@ -1,3 +1,4 @@
 from . import db
 
 enum = db.Enum("A", "B", "C", "D", name="enum")
+simple_entity_enum = db.Enum("A", "B", "C", "D", name="simple_entity_enum")
```
sherryhli, line 4: What's the issue with using `enum` for the simple entity model?

frontend/src/APIClients/SimpleEntityAPIClient.ts
```diff
@@ -36,9 +36,16 @@ const create = async ({
     "accessToken",
   )}`;
   try {
+    // typescript {
     const { data } = await baseAPIClient.post("/simpleEntities", formData, {
       headers: { Authorization: bearerToken },
     });
+    // } typescript
+    // python {
```
sherryhli, line 44: Do you mind changing the route name in the TypeScript backend as well so that we don't need to have 2 separate cases? I think kebab case URLs are much more commonly used than camelCase

backend/python/app/middlewares/validate.py
sherryhli: Got it. Regardless, I think it would be beneficial to document this, in case a different http client is used to make the request :)

sherryhli (commented):
Couple more things

backend/python/app/models/enum.py
alexguo8: I got an error with the migration saying that the type "enum" already exists

sherryhli (dismissed):
LGTM! 🚢

sherryhli (approved):

